PR #241, opened 2026-04-01
docs: add German translation

dknalp:
Added German translation for the README and updated the navigation links in all localized READMEs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 </p>
 
 <p align="center">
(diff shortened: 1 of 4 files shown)


Panniantong:
Thanks for the German translation! There's a merge conflict in README.md (the language nav bar line) after we merged the Korean README (#251). Could you rebase on main and resolve the conflict? It should just be adding both 한국어 and Deutsch to the same nav line. Thanks!

Panniantong:
Thanks for the translation work! Closing under a tightened language policy: only zh/en READMEs are officially maintained — snapshot translations drift fast (this one already teaches the retired `bird` CLI in 8 places and a pre-auth Reddit flow, both changed since April). Rather than ship instructions that would break for German readers, we'll keep zh/en as the sources of truth.

PR #266, opened 2026-04-16
fix: keep install and doctor non-interactive for cookie channels

0xWinner98:
## Summary

Keep `install` and cookie-backed channel health checks non-interactive by default.

## Problem

On macOS, `agent-reach install` and `agent-reach doctor` can route into browser-cookie scanning or interactive status commands for cookie-backed channels.

In practice this can trigger repeated Keychain / password prompts when the underlying CLI scans multiple browsers for cookies. It also makes passive health checks hang or retry in places where the user only asked for install or doctor output.

## Changes

- `install`
  - keep automatic browser cookie import disabled by default
  - add explicit opt-in via `AGENT_REACH_AUTO_IMPORT_COOKIES=1`
  - keep `--dry-run` messaging aligned with the non-interactive default
- `twitter`
  - detect saved local credentials from Agent Reach config, `~/.config/xfetch/session.json`, or `~/.config/bird/credentials.env`
  - skip `twitter status` / `bird check` during doctor-style checks
- `xiaohongshu`
  - detect saved cookies from Agent Reach config or common local cookie files
  - skip `xhs status` during doctor-style checks
- `reddit`
  - detect saved credentials from `~/.config/rdt-cli/credential.json`
  - skip `rdt status --json` during doctor-style checks
- docs
  - document the new install behavior
- tests
  - update channel tests to validate passive local-credential detection
  - add CLI coverage for the non-interactive install messaging

## Why this shape

`install` and `doctor` are passive flows. They work better when they only inspect local state and leave login / browser-cookie import to explicit configuration steps.

That keeps health checks fast, avoids repeated Keychain prompts, and still preserves the existing explicit login/configure workflows.

## Validation

- `python3 -m py_compile agent_reach/channels/twitter.py agent_reach/channels/xiaohongshu.py agent_reach/channels/reddit.py agent_reach/cli.py`
- `.venv/bin/pytest tests/test_twitter_channel.py tests/test_channels.py tests/test_cli.py -q`
- `.venv/bin/pytest tests/test_doctor.py -q`
- `.venv/bin/pytest -q`

Full test run result: `86 passed in 18.83s`

### DIFF
--- a/agent_reach/channels/reddit.py
+++ b/agent_reach/channels/reddit.py
@@ -8,7 +8,7 @@ session cookie. Run `rdt login` after installation to authenticate.
 
 import json
 import shutil
-import subprocess
+from pathlib import Path
 
 from .base import Channel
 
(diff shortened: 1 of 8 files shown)


Panniantong:
Thanks for the thoughtful work — the tests are solid and the server/CI pain point is real. After review we're keeping doctor's semantics as-is for now: its core product value is verified diagnostics, and passive credential-file checks would report ok on expired cookies (silently shifting failures to usage time). The opt-in auto-import idea is good; we may revisit this as an explicit `doctor --online/--offline` flag, and this PR will be a useful reference then. Closing for now — appreciated!

PR #276, opened 2026-04-25
feat(scripts): add Whisper audio transcription with Groq→OpenAI fallback

quietbuildlab:
## Summary

Adds `scripts/transcribe.sh` — a self-contained shell helper that downloads/encodes audio with yt-dlp + ffmpeg and transcribes via Whisper, defaulting to **Groq's free whisper-large-v3** and falling back to **OpenAI's whisper-1** on any HTTP/curl failure.

Complements the existing `YouTubeChannel` by giving agents a way to consume video/audio as text without re-implementing the yt-dlp + ffmpeg + chunking plumbing.

## Why

Groq offers Whisper transcription on a free tier and is significantly faster (LPU). Reaching for a paid OpenAI key for every transcription is unnecessary when Groq handles ~99% of cases — but Groq's free tier does occasionally rate-limit, so an automatic fallback keeps long pipelines reliable.

## Behavior

- **Provider order:** `auto` (default = groq → openai), or pin with `--provider=groq|openai`.
- **Keys:** read from `~/.agent-reach/config.yaml` (`groq_api_key:`, `openai_api_key:`) with shell env taking precedence.
- **No PyYAML dependency** — keys are parsed with awk so the script works on a vanilla python install.
- **Audio prep:** mono / 16kHz / 32kbps m4a (most content stays under the 25MB API limit).
- **Chunking:** files >24MB are split into 10-minute segments via ffmpeg and reassembled.
- **Output:** transcript written to `<output_dir>/transcript.txt` and printed to stdout.

## Usage

```bash
# YouTube URL, defaults to groq with openai fallback
scripts/transcribe.sh https://www.youtube.com/watch?v=...

# Local file, force OpenAI
scripts/transcribe.sh ./recording.m4a /tmp/out --provider=openai

# Help
scripts/transcribe.sh --help
```

## Test plan

- [x] `bash -n scripts/transcribe.sh` — syntax OK
- [x] Local m4a → groq path produces transcript.txt
- [x] YouTube URL → yt-dlp downloads, ffmpeg compresses, transcript returned
- [x] Forced groq path with bad key falls through to openai when `--provider=auto`
- [ ] Reviewer: try `scripts/transcribe.sh --help` and a short YouTube clip end-to-end

## Notes

- Adds no Python deps, no new channel, no doctor changes — purely an optional helper script.
- Happy to rename / move (`tools/`, `agent_reach/skill/scripts/`, etc.) or split into a channel if maintainers prefer.

### DIFF
--- a/scripts/transcribe.sh
+++ b/scripts/transcribe.sh
@@ -1,0 +1,187 @@
+#!/usr/bin/env bash
+# transcribe_audio.sh — Transcribe YouTube/audio via Whisper API.
+#
+# Usage:
+#   transcribe_audio.sh <youtube_url_or_local_file> [output_dir] [--provider=groq|openai|auto]
+#
+# Providers (default order: groq → openai):
+#   groq    — free tier, fast (Groq LPU), whisper-large-v3           [DEFAULT]
+#   openai  — paid (~$0.006/min), reliable fallback, whisper-1
+#   auto    — try groq first, fall back to openai on failure         [DEFAULT]
+#
+# Keys are read from ~/.agent-reach/config.yaml (NOT shell env by default):
+#   groq_api_key:   gsk_...
+#   openai_api_key: sk-...
+#
+# Configure with:
+#   agent-reach configure groq-key   gsk_...
+#   agent-reach configure openai-key sk-...
+#
+# Pipeline:
+#   1. Download audio with yt-dlp (skipped if input is an existing local file).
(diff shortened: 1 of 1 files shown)


quietbuildlab:
Closing in favor of #277, which reshapes this work as a Python module per CONTRIBUTING.md (channel/test/doctor architecture). The new PR also fixes the shell-quoting / chunk-boundary issues called out internally and reuses `agent_reach.config.Config` for key loading instead of a parallel parser.

PR #329, opened 2026-06-08
fix(build): drop duplicate force-include that breaks source installs

nyxst4ck:
## Problem

Every source install of the package is currently broken. Running `python -m build` (or `pip install` from an archive / git URL) fails with:

```
ValueError: A second file is being added to the wheel archive at the same path: `agent_reach/guides/setup-exa.md`.
```

This is the failure reported in #308, #315, #327 and #328. It affects all documented manual install paths:
- `pip install https://github.com/Panniantong/Agent-Reach/archive/main.zip` (README)
- `pipx install .` (docs/install.md)
- `pip install git+https://github.com/Panniantong/Agent-Reach.git`

## Cause

`[tool.hatch.build.targets.wheel.force-include]` re-includes `agent_reach/guides`, `agent_reach/skill` and `agent_reach/scripts` — but those directories already live inside the package and are shipped by `packages = ["agent_reach"]`. Hatchling sees each data file twice and aborts the build (see https://hatch.pypa.io/latest/config/build/#forced-inclusion).

## Fix

Remove the redundant `force-include` table. `packages = ["agent_reach"]` already ships the whole package, data files included.

## Verification (lossless)

After the change, `python -m build --wheel` succeeds and the wheel still contains every data file:

```
agent_reach/guides/setup-exa.md
agent_reach/guides/setup-groq.md
agent_reach/guides/setup-reddit.md
agent_reach/guides/setup-twitter.md
agent_reach/guides/setup-wechat.md
agent_reach/guides/setup-xiaohongshu.md
agent_reach/scripts/transcribe_xiaoyuzhou.sh
agent_reach/skill/SKILL.md
```

`pytest tests/` is unaffected (no runtime code changed).

Fixes #308, fixes #315, fixes #327, fixes #328

PR #331, opened 2026-06-08
test(wechat): add unit tests for WeChatChannel

nyxst4ck:
## What

Adds dedicated unit tests for `WeChatChannel`, the only channel without focused coverage (it was exercised only by the generic contract loop).

Tests cover:
- `can_handle()` — matches `mp.weixin.qq.com` / `weixin.qq.com`, rejects others.
- `check()` — all four backend-detection branches, by toggling `camoufox` and `miku_ai` in `sys.modules`:
  - neither installed → `off`
  - both → `ok` (full)
  - read backend only → `ok` (with search-unlock hint)
  - search backend only → `warn`

## Why

`check()`'s branching is pure and deterministic, so it's cheap to lock down. No network or optional dependencies required — a `None` entry in `sys.modules` forces `ImportError`, a dummy `ModuleType` makes the import succeed, and `monkeypatch` restores both. Follows the existing `TestV2EXChannel` / `TestXueqiuChannel` style.

All 5 new tests pass locally (`pytest tests/test_channels.py::TestWeChatChannel`).

### DIFF
--- a/tests/test_channels.py
+++ b/tests/test_channels.py
@@ -10,6 +10,7 @@ from agent_reach.channels import get_all_channels, get_channel
 from agent_reach.channels.xiaohongshu import XiaoHongShuChannel
 from agent_reach.channels.xueqiu import XueqiuChannel
 from agent_reach.channels.v2ex import V2EXChannel
+from agent_reach.channels.wechat import WeChatChannel
 
 
 class TestChannelRegistry:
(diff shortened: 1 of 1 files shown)


Panniantong:
感谢补测试!两个情况说明:① 实测这组测试是按 v1.4.0 之前的 miku_ai 后端写的,对当前实现 3/5 失败;② 更重要的是,WeChat 渠道刚在 v1.4.2 中整体移除了(#347——全文阅读被反爬拦截越来越严重,与其展示坏掉的承诺不如诚实下架),所以这个测试目标已不存在。你在 #329/#330 的贡献质量很高,欢迎继续参与其他渠道——比如现存渠道的测试覆盖同样欢迎。

nyxst4ck:
Thanks for the clear explanation — that makes total sense. I hadn't caught that the WeChat channel was removed in v1.4.2 (#347), and honestly "honest removal over a broken promise" is the right call. Glad #329 and #330 were useful! I'll take you up on the suggestion and look at test coverage for the current channels next.

PR #346, opened 2026-06-10
chore: stop tracking stale uv.lock (no workflow consumes it)

Panniantong:
## Why

`uv.lock` is a dead file in this repo. Three pieces of evidence:

1. **No workflow consumes it.** The build backend is hatchling (`python -m build`), CI ([.github/workflows/pytest.yml](.github/workflows/pytest.yml)) installs via `pip install -c constraints.txt -e .[dev]`, and `test.sh` also uses pip. A repo-wide grep finds zero occurrences of `uv sync`, `uv run`, or any other consumer of `uv.lock`.
2. **Its contents are stale.** It pins `agent-reach` at `version = "1.3.0"` (last updated in #235), while `pyproject.toml` is already at `1.4.1`. The lockfile has silently drifted from reality.
3. **It only attracts security-bot noise.** e.g. #335 (OrbisAI) bumped a CVE in `uv.lock` alone — a change with zero effect on any actual install path, because nothing ever reads this file.

## What

- `git rm uv.lock`
- Add `uv.lock` to `.gitignore` to prevent accidental re-commits

## Verification

`pytest tests/ -v` — **105 passed** on a clean venv installed the same way CI does (`pip install -c constraints.txt -e .[dev]`).

## Note to maintainer

If you ever want to support a uv workflow in the future, the right move is the opposite of keeping this file: regenerate it with `uv lock` **and** make CI consume it with `uv sync --locked`. Either fully adopt the lockfile or don't track it — an inconsistent lockfile is worse than none, because it advertises guarantees nothing enforces.


Panniantong:
Superseded — uv.lock was removed in #348 (merged), for the same reason described here (no build/CI path consumes it). Closing this as already done.

PR #361, opened 2026-06-12
test(github): add unit tests for GitHubChannel can_handle and check

nyxst4ck:
## What

Adds dedicated offline unit tests for the `github` channel (8 tests, no network and no real `gh` required — `probe_command` is stubbed):

- `can_handle()`: matches on the URL **netloc only** — `github.com` path segments on other hosts (`example.com/github.com/mirror`) must not match
- `check()`: all five branches — `missing` (warn + install link), `broken` install (error + binary reinstall prescription, not the pipx/uv one), status-probe `timeout` (warn, tool considered alive so backend stays active), authenticated (`ok`), and unauthenticated (warn + `gh auth login` — the normal rc≠0 business state, not an error)
- Regression test: a failed `check()` resets a stale `active_backend`

## Why

Follow-up to #331 — pointing the test effort at today's channels, as suggested. The unauthenticated-is-warn-not-error distinction and the timeout-keeps-backend-active behavior are easy to regress silently; these tests pin them. Companion to #360 (rss).

## Verification

```
python -m pytest tests/test_channels.py -k GitHub -v   # 8 passed
python -m pytest tests/                                 # 159 passed, 8 skipped
```

Note: this PR and #360 both append to `tests/test_channels.py`; whichever lands second may need a trivial rebase — happy to do it.

### DIFF
--- a/tests/test_channels.py
+++ b/tests/test_channels.py
@@ -1382,4 +1382,94 @@ class TestRSSChannel:
         monkeypatch.setitem(sys.modules, "feedparser", None)
         status, _ = ch.check()
         assert status == "off"
+
+
+class TestGitHubChannel:
+    """can_handle URL patterns + the five check() branches (missing / broken / timeout / ok / unauthenticated)."""
+
+    def _patch_probe(self, monkeypatch, result):
+        from agent_reach.channels import github as github_mod
+        monkeypatch.setattr(github_mod, "probe_command", lambda *a, **k: result)
+
+    def test_can_handle_github_urls(self):
+        from agent_reach.channels.github import GitHubChannel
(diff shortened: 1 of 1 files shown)


nyxst4ck:
Rebased onto current main to clear the conflict with the RSS-channel tests that landed in the meantime (#360). No content changes — same 8 GitHub-channel tests, still green (249 passed, 11 skipped) on the full suite.

Panniantong:
感谢补充 GitHubChannel 测试。merged #530 已按新的只读 Doctor 语义提供更完整覆盖，而本 PR 仍断言旧的 active_backend 行为并存在同名测试类，因此作为已取代实现关闭。

PR #370, opened 2026-06-14
security: fix yt-dlp arg-injection RCE, SSRF, and install-time bash; mask session cookies

mikerivera33:
Security hardening for three exploitable issues plus two quick wins, all with tests. **No behavior change for normal use** (valid URLs, local-file transcription, and `agent-reach install` on machines that already have Node all work unchanged). A private security advisory with full per-issue repro is being filed separately via SECURITY.md.

- **transcribe.download_audio** — the `source` string (from an LLM or scraped content) was passed to yt-dlp as the final positional with no `--` end-of-options marker, so a value like `--exec=CMD` was parsed as a yt-dlp option (RCE / local-file read). Now validated via new `utils/urlsafe.py` (http(s) scheme allowlist + private/loopback/link-local/metadata-IP block) and a literal `--` is inserted before the URL.
- **channels/web.py** — URL was concatenated raw into the Jina path with no validation, no encoding, uncapped `read()`. Now validated, percent-encoded, capped at 10 MB.
- **cli.py install** — the `curl … setup_22.x` → `bash` step is gated behind `AGENT_REACH_ALLOW_REMOTE_SCRIPTS=1`; default prints trusted manual instructions.
- **config.to_dict** — masks only key/token/password/proxy, so `xhs_cookie`/`xueqiu_cookie`/`bilibili_sessdata` leaked in plaintext. Now masked.
- **CI** — added `permissions: contents: read`.

Tests: 185 passed, 8 skipped; `ruff (E,F,I)` and `mypy` clean on changed/new modules.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -4,6 +4,10 @@ on:
   push:
   pull_request:
 
+# Least privilege: the workflow only needs to read the repo to run tests.
+permissions:
+  contents: read
+
 jobs:
   test:
     runs-on: ubuntu-latest
--- a/agent_reach/channels/web.py
+++ b/agent_reach/channels/web.py
@@ -1,10 +1,16 @@
 # -*- coding: utf-8 -*-
 """Web — any URL via Jina Reader. Always available."""
 
+import urllib.parse
 import urllib.request
+
+from ..utils.urlsafe import assert_safe_public_url, is_http_url
 from .base import Channel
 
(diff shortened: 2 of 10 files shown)


Panniantong:
感谢报告 yt-dlp 参数、SSRF、安装脚本和 session 脱敏风险。这些范围已由 merged #530、#575、#576、#577 分层覆盖并完成组合回归，因此关闭旧大包。

PR #371, opened 2026-06-14
security(channels): match URL host by exact/subdomain, not substring

mikerivera33:
`can_handle` used `"github.com" in urlparse(url).netloc`, which also matches look-alike hosts (`github.com.evil.com`, `evil-github.com`) and userinfo tricks (`github.com@evil.com` — netloc contains the domain but the real host is evil.com). Routing such a URL to a credentialed channel exercises its cookies/session against an unintended host. Adds `utils/urlmatch.host_matches` (parsed hostname, exact-or-subdomain) and switches all 10 domain channels to it. New `tests/test_urlmatch.py`.

### DIFF
--- a/agent_reach/channels/bilibili.py
+++ b/agent_reach/channels/bilibili.py
@@ -39,9 +39,8 @@ class BilibiliChannel(Channel):
     tier = 1
 
     def can_handle(self, url: str) -> bool:
-        from urllib.parse import urlparse
-        d = urlparse(url).netloc.lower()
-        return "bilibili.com" in d or "b23.tv" in d
+        from ..utils.urlmatch import host_matches
+        return host_matches(url, "bilibili.com", "b23.tv")
 
     def check(self, config=None):
         """Probe candidates in order; first fully-usable backend wins."""
--- a/agent_reach/channels/github.py
+++ b/agent_reach/channels/github.py
@@ -13,8 +13,8 @@ class GitHubChannel(Channel):
     tier = 0
 
     def can_handle(self, url: str) -> bool:
-        from urllib.parse import urlparse
-        return "github.com" in urlparse(url).netloc.lower()
+        from ..utils.urlmatch import host_matches
+        return host_matches(url, "github.com")
 
     def check(self, config=None):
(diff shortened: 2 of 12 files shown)


xg-gh-25:
Strong catch on the URL validation flaw. Substring matching in `netloc` is a common pitfall — `urlparse` returns complex structures that require careful extraction.

Your `host_matches` approach is solid. One edge case to consider:

### **Port-number suffix attacks**
`github.com:8080@evil.com` parses with `netloc = "github.com:8080@evil.com"` and `hostname = "evil.com"`. If `host_matches` only checks hostname, it's safe. But if any code path still uses `netloc` directly (e.g., logging, error messages), the port suffix can obfuscate the real host in audit trails.

Ensure **all** references to the extracted domain use `hostname` (post-port-strip), not `netloc`.

### **Subdomain scope clarification**
Your implementation accepts `api.github.com` and `gist.github.com` for a `github.com` channel. This is correct for GitHub's federated services, but for other channels, overly permissive subdomain matching can enable lateral movement (e.g., if `marketing.example.com` is compromised but `api.example.com` holds session tokens).

Consider a per-channel **subdomain whitelist** for sensitive domains:
```python
GITHUB_SUBDOMAINS = {"api", "gist", "raw", "github"}  # explicit allow
```

Then in `host_matches`:
```python
if extracted_subdomain and channel_subdomains:
    if extracted_subdomain not in channel_subdomains:
        return False
```

For most channels, `channel_subdomains=None` → allow any subdomain (current behavior). For sensitive ones, explicit control.

We applied this exact pattern after a security review flagged subdomain wildcarding as an escalation vector.

---
*Security patterns in agent credential routing. Built by [SwarmAI](https://github.com/xg-gh-25/SwarmAI). Discussion: [T-CUL](https://github.com/xg-gh-25/SwarmAI/discussions/8)*

Panniantong:
感谢发现 substring hostname 匹配问题。main 已集中使用 exact-or-subdomain host_matches，并覆盖 lookalike host 与 userinfo 负测，因此关闭。

PR #373, opened 2026-06-14
security(config): create secret directories owner-only (0o700)

mikerivera33:
Config / cookie-sync dirs (`~/.agent-reach`, `~/.config/xfetch`, `~/.config/bird`) hold tokens and session cookies but were created with the default umask (often world-traversable 0o755). Files were already 0o600; this closes the dir gap. Adds `utils.paths.make_private_dir` (mkdir + explicit chmod 0o700; no-op on Windows). New `tests/test_private_dir.py`.

### DIFF
--- a/agent_reach/config.py
+++ b/agent_reach/config.py
@@ -35,8 +35,9 @@ class Config:
         self.load()
 
     def _ensure_dir(self):
-        """Create config directory if it doesn't exist."""
-        self.config_dir.mkdir(parents=True, exist_ok=True)
+        """Create config directory (owner-only; it holds tokens and cookies)."""
+        from agent_reach.utils.paths import make_private_dir
+        make_private_dir(self.config_dir)
 
     def load(self):
         """Load config from YAML file."""
--- a/agent_reach/cookie_extract.py
+++ b/agent_reach/cookie_extract.py
@@ -176,8 +176,9 @@ def _sync_xfetch_session(auth_token: str, ct0: str) -> None:
     import os
 
     try:
+        from agent_reach.utils.paths import make_private_dir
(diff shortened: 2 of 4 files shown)


xg-gh-25:
Clean security hardening — directory permissions are often overlooked until post-incident audit.

Two refinements worth considering:

### 1. **Race window**: `mkdir` → `chmod` sequence
Between `os.makedirs(mode=0o755)` (default umask) and `os.chmod(path, 0o700)`, the directory briefly exists with permissive mode. On multi-user systems, this can leak path names or allow symlink races. Consider:

```python
old_umask = os.umask(0o077)  # Force restrictive creation mask
try:
    os.makedirs(path, mode=0o700, exist_ok=True)
finally:
    os.umask(old_umask)  # Restore original
```

This makes the restrictive mode **atomic at creation time** rather than a two-step fix.

### 2. **Migration path for existing installs**
Users upgrading from older versions have `~/.agent-reach` still at 0o755. The `exist_ok=True` flag means `chmod` won't re-run on existing dirs. Add a one-time migration check in startup (e.g., `stat` → `chmod` if mode != 0o700) or document that users should `chmod 700 ~/.agent-reach` manually.

For reference: we hit this exact pattern when hardening our own agent config dirs — the umask approach eliminated the race window entirely.

---
*Security cultivation in agent tooling. Built by [SwarmAI](https://github.com/xg-gh-25/SwarmAI). Discussion: [T-CUL](https://github.com/xg-gh-25/SwarmAI/discussions/8)*

Panniantong:
感谢补充 secret 目录权限。main 已使用 owner-only 目录、原子写入、symlink refusal 与 no-follow 读取，因此关闭此已覆盖 PR。

PR #375, opened 2026-06-14
security(supply-chain): add hash-pinned runtime lock + document dev-only constraints

mikerivera33:
Every user-facing install path (pip/pipx from main.zip, PyPI) resolved the unbounded `>=` ranges in pyproject and took whatever the index served, with no hashes — a single tampered/yanked-and-republished upstream (yt-dlp, browser-cookie3, requests) reaches all installs. `constraints.txt` only applies to the dev/CI command, giving a false sense of pinning. Adds `requirements.in` (tested versions) + `requirements.lock` (exact versions with SHA-256 hashes) and documents a `--require-hashes` install path; notes `constraints.txt` is dev-only. (Lock generated on Windows/py3.12; installs cross-platform — see header note in requirements.in.)


Panniantong:
感谢关注供应链锁定。当前仓库将 library 兼容范围与测试 constraints 分开维护，本 runtime lock 会引入未验证的平台绑定和额外发布机制，暂不采用并关闭。

Issue #378, opened 2026-06-15
Audit report


Swader:
Agent Reach is not a normal web server, so the main attack surface is not SQL injection or request floods against a public endpoint. The real surface is more dangerous for this product category: local host mutation, supply-chain installation, credential/cookie extraction, subprocess execution, browser profile access, agent-visible docs, and unbounded URL/media processing.

**Production readiness:** not recommended for broad agent/operator use until P0/P1 issues below are fixed.  
**Security posture:** high-risk defaults, mostly from installer behavior and credential handling.  
**DX posture:** docs/API mismatch is severe enough that LLMs and humans will call commands that do not exist.  
**UX posture:** users are likely to be surprised by filesystem mutations, auto-installs, silent fallbacks, and stale command docs.  
**Performance posture:** acceptable for ad hoc CLI usage, but slow/hang-prone for CI/MCP/bot loops.

## Dynamic checks performed:

- Imported channel handlers and tested malicious host routing cases.
- Exercised `Config.to_dict()` redaction behavior with cookie/session fields.
- Injected a fake `rookiepy` module to validate the Xueqiu cookie loading path.
- Ran the test suite with time limits. The suite collected 162 tests and reached roughly 63% progress before timing out in the current environment, around `tests/test_core.py::TestAgentReach::test_doctor_report`. This suggests the suite mixes unit tests with real doctor/probe behavior and lacks strict global test budgets/mocks.

Tooling limitations:

- Local `ruff`, `bandit`, `pip-audit`, `semgrep`, and `mypy` were not installed in the container, so this report is primarily manual plus targeted dynamic repros.
- I did a dependency freshness/vulnerability spot check separately; see the dependency section.

## Severity model

- **P0 / Critical:** likely host compromise, credential compromise, large trust violation, or default behavior that can break user machines.
- **P1 / High:** exploitable security gap, costly privacy/cost/resource surprise, major agent/product failure, or common path denial-of-service.
- **P2 / Medium:** important correctness, reliability, hardening, or maintainability issue.
- **P3 / Low:** cleanup, polish, weak signals, or future-proofing.

---

# Findings

## P0-01 — Installer performs high-risk host mutations and unpinned third-party installs by default

**Owners:** Security, DX, UX, Team Lead  
**Files:** `agent_reach/cli.py`  
**Evidence:**

- `install --safe` exists, but safe mode is opt-in (`cli.py:71-74`).
[...]

PR #389, opened 2026-06-18
fix(core): use PEP 440 version comparison and fix Xueqiu cookie loading

StevanusPangau:
## Summary

Two bugfixes from audit report #378:

### Fix 1: P3-15 — Version comparison (`cli.py`)

**Problem:** `_is_newer_version()` uses manual tuple parsing that cannot handle PEP 440 versions (pre-release, post-release, dev, build metadata).

**Fix:** Use `packaging.version.Version` for proper PEP 440 comparison with graceful fallback to tuple comparison if parsing fails.

### Fix 2: P2-08 — Xueqiu browser cookie loading (`channels/xueqiu.py`)

**Problem:** `_cookie_jar.set(name, value, domain=...)` is called on `http.cookiejar.CookieJar` which does **not** have a `.set()` method (only `requests.cookies.RequestsCookieJar` does). This causes a silent `AttributeError` caught by the broad `except Exception`, making the rookiepy cookie path always fail.

**Fix:** Construct proper `http.cookiejar.Cookie` objects and use `_cookie_jar.set_cookie()`, consistent with the existing `_inject_cookie_string()` pattern in the same file.

## Changes

- `agent_reach/cli.py` — Replaced manual version parsing with `packaging.version.Version`
- `agent_reach/channels/xueqiu.py` — Fixed cookie loading to use proper Cookie objects
- `pyproject.toml` — Added `packaging>=21.0` dependency

## Verification

- `pip install -e .` — success
- `pytest tests/ -v` — **162 tests passed**

## References

Addresses P3-15 and P2-08 from #378

### DIFF
--- a/agent_reach/channels/xueqiu.py
+++ b/agent_reach/channels/xueqiu.py
@@ -85,7 +85,25 @@ def _load_cookies_from_browser() -> bool:
             if not any(c.get("name") == "xq_a_token" for c in cookies):
                 return False
             for c in cookies:
-                _cookie_jar.set(c["name"], c["value"], domain=c.get("domain", ".xueqiu.com"))
+                cookie = http.cookiejar.Cookie(
+                    version=0,
+                    name=c["name"],
+                    value=c["value"],
+                    port=None,
+                    port_specified=False,
+                    domain=c.get("domain", ".xueqiu.com"),
+                    domain_specified=True,
+                    domain_initial_dot=True,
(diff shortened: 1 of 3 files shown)


Panniantong:
感谢修复版本比较和雪球 Cookie 路径。main 已采用当前版本检查语义并修复雪球会话与估值路径，本旧组合 PR 已冲突且被覆盖，因此关闭。

PR #405, opened 2026-06-21
feat(health): add Channel Health Registry

wby1121:
Lightweight in-memory health registry populated transparently when doctor runs.  Exposes current channel state (HEALTHY/DEGRADED/UNAVAILABLE) via gent-reach health and the get_registry() API.

- ChannelStatus, ChannelHealth, HealthRegistry in new health.py
- doctor.check_all() stores results in the shared registry
- gent-reach health prints a simple status table
- No routing, scoring, or persistence — just state

PR #406, opened 2026-06-22
perf(doctor): probe channels concurrently instead of sequentially

An-idd:
## Summary

`doctor` probes every channel **sequentially** today ([`check_all`](https://github.com/Panniantong/Agent-Reach/blob/main/agent_reach/doctor.py)), and each `check()` does real subprocess/network probing with multi-second timeouts (e.g. `rdt status` waits up to 10s, `opencli daemon status`, `gh`/bilibili network calls). So the doctor's wall time is the **sum** of every channel's probe time.

Probing is I/O-bound, so a `ThreadPoolExecutor` sidesteps the GIL and collapses wall time to roughly the **single slowest channel**.

## Before / After

Measured locally (most upstream tools *not* installed, so only the network-touching channels are slow):

| | wall time |
|---|---|
| Sequential (sum of probes) | **2.63s** |
| Parallel (slowest single channel) | **1.08s** (~2.4× faster) |

Slowest channels here: `github` (1.08s), `bilibili` (1.04s), `v2ex` (0.43s) — all real network I/O.

The win grows in real environments: with `rdt-cli` installed, the Reddit probe alone can block up to 10s, stacked on top of the OpenCLI daemon probes (Reddit / Twitter / Bilibili / Xiaohongshu). Sequentially that is 20–30s+; in parallel it is ~the slowest single probe.

## Why it's safe

- Each channel is a **distinct registry object**; the shared probe helpers (`opencli_status`, `probe_command`) are **stateless** (verified: no file writes or shared mutable state on any `check()` path).
- Per-channel `try/except` is preserved in `_check_one`, which **never raises**, so one bad channel still degrades to `status="error"` exactly as before.
- Result order still follows the channel registry — `ThreadPoolExecutor.map()` preserves input order — which `format_report()` relies on for tiered output.

## Testing

- New test `test_check_all_runs_channels_concurrently`: 8 stub channels each sleeping 0.3s (2.4s sequential) must finish well under 1s **and** preserve registry order.
- Full suite green: **163 passed**.
- Verified real `agent-reach doctor` output is unchanged (same tiers, same ordering, same active-backend annotations).

## Note on overlap with #405

#405 (Channel Health Registry) also edits `check_all` / `tests/test_doctor.py`, so there may be a textual merge conflict. They are **semantically compatible**: this PR keeps `check_all` producing the identical `results` dict, so #405's registry population (`update_from_doctor_results(results)` before `return`) slots in with a one-line rebase whichever lands first.

### DIFF
--- a/agent_reach/doctor.py
+++ b/agent_reach/doctor.py
@@ -4,35 +4,54 @@
 Each channel knows how to check itself. Doctor just collects the results.
 """
 
-from typing import Dict
+from concurrent.futures import ThreadPoolExecutor
+from typing import Dict, Tuple
 from agent_reach.config import Config
-from agent_reach.channels import get_all_channels
+from agent_reach.channels import Channel, get_all_channels
 
 
-def check_all(config: Config) -> Dict[str, dict]:
-    """Check all channels and return status dict.
+def _check_one(ch: Channel, config: Config) -> Tuple[str, dict]:
+    """Probe a single channel; return (name, result). Never raises.
 
     A single misbehaving channel must never take the whole report down,
     so per-channel exceptions degrade to status="error".
     """
-    results = {}
-    for ch in get_all_channels():
-        try:
-            status, message = ch.check(config)
-            active = getattr(ch, "active_backend", None)
-        except Exception as e:  # noqa: BLE001 — doctor must survive any channel
-            # Channels are registry singletons: a stale active_backend from a
-            # previous check must not leak into an errored result.
-            status, message, active = "error", f"体检异常：{e}", None
-        results[ch.name] = {
-            "status": status,
-            "name": ch.description,
-            "message": message,
-            "tier": ch.tier,
-            "backends": ch.backends,
-            "active_backend": active,
-        }
-    return results
+    try:
+        status, message = ch.check(config)
+        active = getattr(ch, "active_backend", None)
+    except Exception as e:  # noqa: BLE001 — doctor must survive any channel
+        # Channels are registry singletons: a stale active_backend from a
+        # previous check must not leak into an errored result.
+        status, message, active = "error", f"体检异常：{e}", None
+    return ch.name, {
+        "status": status,
+        "name": ch.description,
+        "message": message,
+        "tier": ch.tier,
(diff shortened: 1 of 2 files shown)


An-idd:
### Real-environment benchmark (installed tools)

Fair A/B on an **identical installed state**: sequential = looping `ch.check()` in registry order (the old behavior); parallel = the new `check_all()`. Medians over 3 trials, raw trials shown for transparency (network probes are inherently noisy).

**1. Light env** (no upstream tools installed — only network-probing channels are slow):

| | wall time |
|---|---|
| Sequential | **4.1s** (trials: 4.03, 4.09, 4.71) |
| Parallel | **1.8s** (trials: 1.50, 1.84, 4.56) |

→ ~2.2×

**2. With a real backend installed** (`pip install rdt-cli`, not logged in → the Reddit probe hits its 10s timeout cap, which is exactly what real users hit):

| | wall time |
|---|---|
| Sequential | **18.0s** (trials: 6.53, 17.97, 38.1) |
| Parallel | **5.0s** (trials: 3.81, 4.98, 10.02) |

→ ~3.6×

Per-channel probe breakdown from that state makes the mechanism obvious:

```
   10008 ms  reddit      <- rdt status timeout cap
    1013 ms  xueqiu
     950 ms  github
     652 ms  bilibili
     422 ms  v2ex
      <5 ms  everything else
  -------------------------------
   SUM (sequential)  = 13.1s
   MAX (parallel ~=) = 10.0s
```

**The invariant:** sequential wall = **Σ**(probe times); parallel wall ≈ **max**(probe times). The win scales with the number of *concurrently-slow* channels. On a real desktop where OpenCLI backs Reddit / Twitter / Bilibili / Xiaohongshu, several probes are slow at once, so the gap widens further toward N×.

Panniantong:
Doctor 有界并发的性能价值成立，但当前旧补丁会丢失 main 的 URL 凭据清洗，并忽略全局 channel 单例在并发 Doctor 调用中的 active_backend 竞态；墙钟测试也不稳定，因此关闭并等待从最新 main 重新实现。

PR #416, opened 2026-06-23
test(cli): cover XHS Docker container detection

### LINKED ISSUE #410: [BUG] `agent-reach configure xhs-cookies` always fails: `docker ps --format ""` always returns empty string → container permanently detected as "not running"
## Summary

`agent-reach configure xhs-cookies` **always** prints `"[X] xiaohongshu-mcp container is not running"` and exits early, even when the container is actually running. The entire cookie-write flow (docker cp + container restart + mcporter verification) is **completely unreachable** due to a one-line bug in the `docker ps` invocation.

## Root Cause

**File:** `agent_reach/cli.py`, function `_configure_xhs_cookies()`, approximately line 1081

**Buggy code:**
```python
result = subprocess.run(
    [docker, "ps", "--filter", "name=xiaohongshu-mcp", "--format", ""],
    capture_output=True, encoding="utf-8", timeout=5,
)
container_name = result.stdout.strip()
if not container_name:
    print("[X] xiaohongshu-mcp container is not running.")
    ...
    return
```

**The problem:** `docker ps --format ""` (empty format string) always produces **empty output**, regardless of whether any containers are running. Docker interprets an empty template as "print nothing for each row". Therefore `result.stdout` is always `""`, `container_name` is always `""`, and `if not container_name:` is always `True`.

**Reproduce in 2 seconds:**
```bash
# Start the container
docker run -d --name xiaohongshu-mcp -p 18060:18060 xpzouying/xiaohongshu-mcp

# Verify it IS running
docker ps --filter name=xiaohongshu-mcp
# → Shows the container row ✅

# Reproduce the bug: same command with --format ""
docker ps --filter name=xiaohongshu-mcp --format ""
# → (empty output) ← BUG: always empty
```

Running `agent-reach configure xhs-cookies '...'` with the container running always hits the early-return branch and never reaches the `docker cp` / `docker restart` / mcporter verification steps.

## Impact

- **All users** of the `xhs-cookies` configure command are affected on every platform (macOS/Linux/Windows with Docker).
- There is **no workaround** within the tool — users cannot inject XHS cookies through `agent-reach configure xhs-cookies` at all.
- The feature has been broken since it was introduced.

## Fix

Change the empty format string to the Go template that returns the container name:

```diff
- result = subprocess.run(
-     [docker, "ps", "--filter", "name=xiaohongshu-mcp", "--format", ""],
-     capture_output=True, encoding="utf-8", timeout=5,
- )
+ result = subprocess.run(
+     [docker, "ps", "--filter", "name=xiaohongshu-mcp", "--format", "{{.Names}}"],
+     capture_output=True, encoding="utf-8", timeout=5,
+ )
```

`docker ps --format "{{.Names}}"` correctly outputs the container name (one per line) when the container is running, and empty output when it is not — exactly the semantics the surrounding code expects.

## Environment

- **Affected versions:** All (bug present since `_configure_xhs_cookies` was introduced)
- **OS:** macOS / Linux / Windows (any system with Docker)
- **Reproducible:** 100% — no conditions required other than having Docker installed

snvtac:
## Summary
- Add regression coverage for the `configure xhs-cookies` Docker container detection path.
- Assert `docker ps` uses `--format "{{.Names}}"`, so it continues to emit running container names instead of an empty template.
- Verify the running-container flow reaches `docker cp` and does not incorrectly report `xiaohongshu-mcp` as stopped.

Closes #410

## Testing
- `python3 -m py_compile tests/test_cli.py agent_reach/cli.py`
- Manual stdlib simulation of `_configure_xhs_cookies("a=1; b=2")` with mocked Docker commands passed.
- Not run: `python3 -m pytest tests/test_cli.py -q` (local `python3` has no `pytest`; did not install dependencies per operator instruction).
- Not run: `python3 -m ruff check tests/test_cli.py agent_reach/cli.py` and `python3 -m ruff format --check tests/test_cli.py` (local `python3` has no `ruff`; did not install dependencies per operator instruction).

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -60,6 +60,44 @@ class TestCLI:
         assert auth_token == "token123"
         assert ct0 == "ct0abc"
 
+    def test_configure_xhs_cookies_detects_running_docker_container(self, monkeypatch, capsys):
+        calls = []
+
+        def fake_which(name):
+            if name == "docker":
+                return "/usr/bin/docker"
+            return None
+
+        def fake_run(cmd, **kwargs):
+            calls.append(cmd)
+            if cmd[:2] == ["/usr/bin/docker", "ps"]:
+                return subprocess.CompletedProcess(cmd, 0, "xiaohongshu-mcp\n", "")
+            if cmd[:4] == ["/usr/bin/docker", "exec", "xiaohongshu-mcp", "printenv"]:
+                return subprocess.CompletedProcess(cmd, 0, "/app/data/cookies.json\n", "")
+            if cmd[:2] == ["/usr/bin/docker", "cp"]:
+                return subprocess.CompletedProcess(cmd, 0, "", "")
+            if cmd == ["/usr/bin/docker", "restart", "xiaohongshu-mcp"]:
+                return subprocess.CompletedProcess(cmd, 0, "xiaohongshu-mcp\n", "")
+            raise AssertionError(f"unexpected docker command: {cmd}")
+
+        monkeypatch.setattr(shutil, "which", fake_which)
+        monkeypatch.setattr(subprocess, "run", fake_run)
+
+        cli._configure_xhs_cookies("a=1; b=2")
+
+        out = capsys.readouterr().out
+        assert calls[0] == [
+            "/usr/bin/docker",
+            "ps",
+            "--filter",
+            "name=xiaohongshu-mcp",
+            "--format",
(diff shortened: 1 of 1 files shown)


Panniantong:
感谢补充小红书 Docker 检测测试。main 已覆盖容器发现、Cookie 临时文件清理和明确状态语义，本旧测试基线已被取代，因此关闭。

PR #418, opened 2026-06-23
fix(cli): preserve skill customizations during doctor

### LINKED ISSUE #413: [BUG] `agent-reach doctor` silently overwrites user-customized SKILL.md on every run via unconditional `_install_skill()` call
## Summary

Running `agent-reach doctor` calls `_install_skill()` **unconditionally on every execution**. This silently overwrites any user customizations made to `SKILL.md` (the Agent skill prompt file) without warning, confirmation, or backup. Users who have tuned their skill file to fit their workflow lose all customizations every time they run a health check.

## Root Cause

**File:** `agent_reach/cli.py`, function `_cmd_doctor()`

**Buggy code:**
```python
def _cmd_doctor(args=None):
    from agent_reach.config import Config
    from agent_reach.doctor import check_all, format_report
    try:
        from rich import print as rprint
    except ImportError:
        rprint = print
    config = Config()
    results = check_all(config)
    if args is not None and getattr(args, "json", False):
        print(json.dumps(results, ensure_ascii=False, indent=2))
        return
    rprint(format_report(results))
    # Auto-install skill if not already present (fixes #154)
    _install_skill()   # ← called EVERY TIME, comment says "if not already present" but code doesn't check
```

The inline comment `# Auto-install skill if not already present (fixes #154)` **contradicts the actual behavior**: `_install_skill()` is called unconditionally with no existence check. Whatever `_install_skill()` does, it runs and overwrites on **every single `doctor` invocation**.

## Steps to Reproduce

```bash
# 1. Install Agent Reach and run doctor once
agent-reach doctor

# 2. Customize your skill file
echo "# My custom instructions for this project" >> ~/.agents/skills/agent-reach/SKILL.md
# (or edit ~/.claude/skills/agent-reach/SKILL.md)

# 3. Run doctor again (e.g. to check if yt-dlp is working)
agent-reach doctor

# 4. Check your customizations — they are gone
cat ~/.agents/skills/agent-reach/SKILL.md
# → Only the default upstream SKILL.md content remains
```

## Impact

- **UX regression:** `agent-reach doctor` is a diagnostic command users run frequently. It is not expected to modify files.
- **Data loss:** User customizations to the Agent skill prompt are silently destroyed with no warning, no backup, and no `--force` opt-in.
- **Misleading comment:** The `# Auto-install skill if not already present` comment creates false confidence that the check is safe to run.

## Fix

Add an existence check before calling `_install_skill()`, matching the intent of the comment:

```diff
 rprint(format_report(results))
-# Auto-install skill if not already present (fixes #154)
-_install_skill()
+# Auto-install skill only if not already present (fixes #154)
+skill_paths = [
+    os.path.expanduser("~/.agents/skills/agent-reach/SKILL.md"),
+    os.path.expanduser("~/.claude/skills/agent-reach/SKILL.md"),
+    os.path.expanduser("~/.openclaw/skills/agent-reach/SKILL.md"),
+]
+if not any(os.path.exists(p) for p in skill_paths):
+    _install_skill()
```

Or, if `_install_skill()` already contains the logic to determine the correct path, expose a `force=False` parameter and pass `force=False` from `_cmd_doctor()`.

## Environment

- **Affected versions:** All (regression since `_install_skill()` was added to `_cmd_doctor()`)
- **OS:** macOS / Linux / Windows
- **Reproducible:** 100% — any `agent-reach doctor` call after customizing SKILL.md

snvtac:
## Summary

- Make `agent-reach doctor` install missing Agent Reach skills without overwriting existing `SKILL.md` files.
- Keep explicit skill installation behavior unchanged by preserving overwrite-by-default for `_install_skill()`.
- Add regression coverage for preserving customized skills, backfilling missing skill directories, and avoiding misleading install output for skipped targets.

## Root Cause

`agent-reach doctor` called `_install_skill()` unconditionally after printing the health report. `_install_skill()` clears and recreates existing skill directories, so a diagnostic command could silently replace a user-customized `SKILL.md` on every run.

## Validation

- `python -m pytest tests/test_cli.py::TestCLI::test_doctor_preserves_existing_skill_install tests/test_cli.py::TestCLI::test_doctor_installs_skill_when_missing tests/test_cli.py::TestCLI::test_doctor_runs`
- `python -m pytest`
- `python -m ruff check tests/test_cli.py`
- `git diff --check`

Notes: full-project `mypy agent_reach/cli.py` and `ruff check agent_reach tests` still report pre-existing issues outside this change, while the new regression path and full test suite pass.

Fixes #413

### DIFF
--- a/agent_reach/cli.py
+++ b/agent_reach/cli.py
@@ -9,10 +9,10 @@ Usage:
     agent-reach setup
 """
 
-import sys
 import argparse
 import json
 import os
+import sys
 import time
 
 from agent_reach import __version__
@@ -340,11 +340,25 @@ def _cmd_install(args):
         print("Dry run complete. No changes were made.")
 
 
-def _install_skill():
+def _skill_parent_dirs():
+    """Return known agent skill parent directories in installation priority order."""
+    skill_dirs = [
+        os.path.expanduser("~/.agents/skills"),      # Generic agents (priority)
+        os.path.expanduser("~/.openclaw/skills"),    # OpenClaw
+        os.path.expanduser("~/.claude/skills"),      # Claude Code (if exists)
+    ]
+
+    openclaw_home = os.environ.get("OPENCLAW_HOME")
+    if openclaw_home:
+        skill_dirs.insert(0, os.path.join(openclaw_home, ".openclaw", "skills"))
+
+    return skill_dirs
+
+
+def _install_skill(overwrite_existing: bool = True):
     """Install Agent Reach as an agent skill (OpenClaw / Claude Code / .agents)."""
-    import os
-    import shutil
     import importlib.resources
+    import shutil
 
     def _is_english_locale(value: str) -> bool:
         normalized = value.strip().lower()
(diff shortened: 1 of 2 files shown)


Panniantong:
Thanks for the contribution. #448 now preserves existing skill customizations during doctor, fixes the misleading output, updates docs, and includes tests. Closing this PR as superseded by the merged implementation.

PR #434, opened 2026-06-26
fix(doctor): stop _cmd_doctor from overwriting user-customized SKILL.md

### LINKED ISSUE #413: [BUG] `agent-reach doctor` silently overwrites user-customized SKILL.md on every run via unconditional `_install_skill()` call

singlaamitesh:
## Description

`agent-reach doctor` silently overwrites user-customized skill files on every run. When `_cmd_doctor()` is called, it ends with `_install_skill()` — which clears and rewrites the entire skill directory (including SKILL.md) without any existence check. Users who tuned their agent prompt lose all customizations each time they run diagnostics.

Closes #413

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changes Made

**`agent_reach/cli.py`:**
- Added `force: bool = True` parameter to `_install_skill()`
- When `force=False`: skip installation if skill directory with SKILL.md already exists
- When `force=True` (default): original behavior — clear and rewrite
- `_cmd_doctor()` now calls `_install_skill(force=False)` to preserve user customizations
- `_copy_skill_dir()` accepts `force` param — only clears existing when `force=True`

**`tests/test_cli.py`:**
- Added `test_install_skill_force_false_skips_overwrite` — verifies custom SKILL.md is preserved
- Added `test_install_skill_force_true_overwrites` — verifies force mode still works

## Testing

- [x] Unit tests pass (`pytest`)
- [x] Linting passes (`ruff check .` + `ruff format`)
- [x] New tests added for this fix
- [ ] Manual testing performed

### Test Output

```text
$ python3 -m pytest tests/test_cli.py::test_install_skill_force_false_skips_overwrite tests/test_cli.py::test_install_skill_force_true_overwrites -xvs
2 passed in 0.03s
```

## Real Behavior Proof

- Environment: macOS 26.5.1 (aarch64), Python 3.14
- Exact command / steps: create `~/.agents/skills/agent-reach/SKILL.md` with custom content, run `agent-reach doctor`
- Observed result (before fix): custom content overwritten with default SKILL.md
- Observed result (after fix): custom content preserved, doctor runs diagnostics normally
- Not tested: Windows/Linux runtime (verified via unit test with tmp_path)

## Review Readiness

- [x] I have performed a self-review
- [x] This PR is ready for human review

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective
- [x] New and existing unit tests pass locally

### DIFF
--- a/agent_reach/cli.py
+++ b/agent_reach/cli.py
@@ -18,7 +18,9 @@ import time
 from agent_reach import __version__
 
 # Pinned to the 0.4.2 state — PyPI still only has 0.4.1 (upstream issue #10).
-_RDT_GIT_SOURCE = "git+https://github.com/public-clis/rdt-cli.git@5e4fb3720d5c174e976cd425ccc3b879d52cac66"
+_RDT_GIT_SOURCE = (
+    "git+https://github.com/public-clis/rdt-cli.git@5e4fb3720d5c174e976cd425ccc3b879d52cac66"
+)
 
 
 def _ensure_utf8_console():
@@ -30,6 +32,7 @@ def _ensure_utf8_console():
         return
     try:
         import io
+
         if hasattr(sys.stdout, "buffer"):
             sys.stdout = io.TextIOWrapper(sys.stdout.buffer, encoding="utf-8", errors="replace")
         if hasattr(sys.stderr, "buffer"):
@@ -42,6 +45,7 @@ def _ensure_utf8_console():
 def _configure_logging(verbose: bool = False):
     """Suppress loguru output unless --verbose is set."""
     from loguru import logger
+
     logger.remove()  # Remove default stderr handler
     if verbose:
         logger.add(sys.stderr, level="INFO")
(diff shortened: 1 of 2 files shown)


Panniantong:
Thanks for the fix. #448 now implements the non-destructive doctor behavior, updates the user-facing output/docs, and adds a regression test. Closing this PR as superseded by the merged implementation.

PR #436, opened 2026-06-26
fix(deps): bump yt-dlp, requests, python-dotenv to patch disclosed CVEs

aeonframework:
Hi — thanks for Agent Reach! This is a small, automated dependency-bump PR that updates three pinned packages in `constraints.txt` to releases that fix **already-public** CVEs. No source changes; only the tested-version pins move.

### yt-dlp `2025.5.22` → `2026.6.9`
This is the meaningful one — `agent_reach/transcribe.py` feeds remote URLs to `yt-dlp`, so the library is the most exposed dependency. `2026.6.9` fixes five advisories:

| Advisory | Severity | Issue |
|---|---|---|
| [GHSA-69qj-pvh9-c5wg](https://github.com/yt-dlp/yt-dlp/security/advisories/GHSA-69qj-pvh9-c5wg) | High | Command injection via `--exec` output template |
| [GHSA-vx4q-3cr2-7cg2](https://github.com/yt-dlp/yt-dlp/security/advisories/GHSA-vx4q-3cr2-7cg2) / CVE-2026-50574 | High | Code execution via manifest downloads with aria2c |
| [GHSA-c6mh-fpjc-4pr3](https://github.com/yt-dlp/yt-dlp/security/advisories/GHSA-c6mh-fpjc-4pr3) / CVE-2026-50023 | High | Dangerous file-type creation (`.desktop`/`.url`/`.webloc`) |
| [GHSA-g3gw-q23r-pgqm](https://github.com/yt-dlp/yt-dlp/security/advisories/GHSA-g3gw-q23r-pgqm) / CVE-2026-26331 | High | Command injection via `--netrc-cmd` |
| [GHSA-f7j3-774f-rfhj](https://github.com/yt-dlp/yt-dlp/security/advisories/GHSA-f7j3-774f-rfhj) / CVE-2026-50019 | Moderate | Cookie leak via the `curl` external downloader |

### requests `2.32.5` → `2.33.0`
[GHSA-gc5v-m9x4-r6x2](https://github.com/psf/requests/security/advisories/GHSA-gc5v-m9x4-r6x2) / CVE-2026-25645 (Moderate) — insecure temp-file reuse in `extract_zipped_paths()`. Not reachable from Agent Reach's code (that helper isn't called), so this is a defense-in-depth bump.

### python-dotenv `1.2.1` → `1.2.2`
[...]

PR #407, opened 2026-06-22
fix(imports): sort import statements according to ruff I001

hsms4710-pixel:
## Summary

Fixes import statement sorting and formatting issues (ruff I001) without changing any business logic.

## Changes

- Fixed 17 import sorting issues (I001)
- Modified 11 files
- Used `ruff --select I001 --fix` for automatic fixing

## Safety

- No imports were removed (preserved for potential future use)
- No business logic was changed
- No functional code was modified

## Verification

- ruff I001 check passes (0 errors)
- Only code style changes, no functional impact

## Why This Matters

Consistent import sorting improves code readability and prevents merge conflicts when multiple developers modify the same file.

PR #437, opened 2026-06-27
fix(meta): add Python 3.13 classifier, fix yt-dlp specifier, bump pytest for 3.13

littletrain-jyp:
## Problem

Three metadata/packaging bugs that affect PyPI discoverability, installation, and CI reliability.

---

### 1 — Python 3.13 missing from classifiers

The CI matrix already tests Python 3.10 – 3.13 (all four), but `pyproject.toml` only declares classifiers through 3.12. Users browsing PyPI see an unverified 3.13 combination and may avoid installing on 3.13.

```toml
# Added
"Programming Language :: Python :: 3.13",
```

---

### 2 — `yt-dlp>=2024.0` references a non-existent release

yt-dlp uses **date-based versioning** (`YYYY.M.D`). The release `2024.0` was never published — the first 2024 release was `2024.1.1`.  
While pip resolves the constraint correctly in practice (it finds the next higher real version), the specifier is misleading and can confuse tooling that reads the metadata.

```toml
# Before
"yt-dlp>=2024.0",
# After
"yt-dlp>=2024.1.1",
```

---

### 3 — `pytest==8.0.0` predates Python 3.13 in `constraints.txt`

pytest 8.0.0 was released 2024-01-27, **before** Python 3.13 went GA (2024-10-07). It is not in pytest's own tested matrix for 3.13. The 3.13 CI leg therefore runs against an untested pytest version.

- Pinned to `pytest==8.4.2` (latest stable, fully supports 3.13).
- Raised `pyproject.toml` dev lower bound from `>=8.0` → `>=8.1` (the first release that began tracking 3.13 compatibility).

```
# constraints.txt
- pytest==8.0.0
+ pytest==8.4.2
```

---

## Files changed

| File | Change |
|------|--------|
| `pyproject.toml` | Add 3.13 classifier; fix yt-dlp specifier; raise pytest lower bound |
| `constraints.txt` | Pin `pytest==8.4.2` |

## Relation to existing PRs

Checked against all open PRs (#407–#436) — none touches classifiers, yt-dlp version specifiers, or the pytest pin in `constraints.txt`.

### DIFF
--- a/constraints.txt
+++ b/constraints.txt
@@ -10,7 +10,7 @@ PyYAML==6.0.3
 rich==14.3.2
 yt-dlp==2025.5.22
 
-pytest==8.0.0
+pytest==8.4.2
 ruff==0.15.1
 mypy==1.19.1
 types-requests==2.32.4.20260107
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,6 +23,7 @@ classifiers = [
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
     "Topic :: Internet :: WWW/HTTP",
     "Topic :: Scientific/Engineering :: Artificial Intelligence",
 ]
@@ -34,7 +35,7 @@ dependencies = [
     "loguru>=0.7",
     "pyyaml>=6.0",
     "rich>=13.0",
-    "yt-dlp>=2024.0",
+    "yt-dlp>=2024.1.1",
 ]
 
 [project.optional-dependencies]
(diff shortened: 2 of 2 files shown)


littletrain-jyp:
Closing this PR — on reflection the changes don't hold up:

- **yt-dlp specifier** ( → ): purely cosmetic. pip's version ordering means these two specifiers resolve identically in practice ( is True), so no install behavior changes.
- **pytest bump** ( → ): the premise that pytest 8.0.0 is incompatible with Python 3.13 is not verified. The CI matrix already tests 3.13 with the 8.0.0 pin and there's no evidence of failures.
- **Python 3.13 classifier**: this change is correct, but too minor to justify a standalone PR — better included as part of a broader metadata update.

Leaving #438 open, which fixes a confirmed runtime bug.

PR #458, opened 2026-07-01
fix: pin mcporter config add to --scope home

HairtonicDublin:
## Summary

`mcporter config add` defaults to `--scope project`, which resolves relative to the current working directory rather than a stable global location. Every call site in Agent Reach that runs this command (the `install --env=auto` flow, the interactive `setup` wizard, and all the doctor-hint / documentation strings for Exa, XiaoHongShu, and LinkedIn) omits an explicit scope.

Practical effect: a server registered during `agent-reach install` (e.g. Exa search) only resolves correctly while mcporter is invoked from the exact directory the installer happened to run in. Run `agent-reach doctor` from a different project folder, or in a fresh terminal/session elsewhere, and `mcporter config list` no longer finds the entry — the channel silently reports as unconfigured even though install previously reported success.

This is easy to hit by accident: I ran into it because I'd cloned Agent Reach's own source into my working directory, and the repo happens to ship an example `config/mcporter.json` fixture — `mcporter config add` silently wrote into that tracked file instead of a real global config, since project-scope resolution found it first. But the same failure mode applies to any user who simply runs `agent-reach install` from one folder and later uses the tool from another.

## Fix

Add `--scope home` to every `mcporter config add` invocation and instruction string, so registrations always land in `~/.mcporter/mcporter.json` regardless of cwd:

- `agent_reach/cli.py` — both the automated install path and the interactive `setup` wizard's actual `subprocess.run` calls, plus all printed manual-fallback instructions
- `agent_reach/channels/exa_search.py`, `xiaohongshu.py`, `linkedin.py` — doctor hint strings shown to users/agents when a channel isn't configured
- `docs/install.md`, `agent_reach/guides/setup-exa.md`, `agent_reach/guides/setup-xiaohongshu.md` — copy-pasteable instructions

## Test plan

- [x] `pytest tests/` — 196 passed, no regressions (the one existing assertion on this string, `tests/test_channels.py:1018`, is a substring check and still holds)
- [x] `ruff check` on all touched files — no new lint errors introduced (pre-existing unrelated lint debt elsewhere in `cli.py` untouched)
- [x] Manually verified: `mcporter config add exa https://mcp.exa.ai/mcp --scope home` writes to `~/.mcporter/mcporter.json`, and `mcporter config list --json` resolves it correctly from an unrelated directory (previously returned empty)

### DIFF
--- a/agent_reach/channels/exa_search.py
+++ b/agent_reach/channels/exa_search.py
@@ -25,7 +25,7 @@ class ExaSearchChannel(Channel):
             return "off", (
                 "需要 mcporter + Exa MCP。安装：\n"
                 "  npm install -g mcporter\n"
-                "  mcporter config add exa https://mcp.exa.ai/mcp"
+                "  mcporter config add exa https://mcp.exa.ai/mcp --scope home"
             )
         if probe.status == "broken":
             return "error", _MCPORTER_BROKEN_HINT
@@ -36,5 +36,5 @@ class ExaSearchChannel(Channel):
             return "ok", "全网语义搜索可用（免费，无需 API Key）"
         return "off", (
             "mcporter 已装但 Exa 未配置。运行：\n"
-            "  mcporter config add exa https://mcp.exa.ai/mcp"
+            "  mcporter config add exa https://mcp.exa.ai/mcp --scope home"
         )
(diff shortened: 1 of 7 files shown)


Panniantong:
感谢指出 mcporter 配置作用域。main 的安装器、Doctor 提示和文档均已统一使用 --scope home，因此关闭此已覆盖 PR。

PR #459, opened 2026-07-01
fix(youtube): harden yt-dlp against bot-check and JS-challenge failures

HairtonicDublin:
## Summary

Testing the documented YouTube commands against a real video surfaced two compounding failure modes not covered by the current docs:

1. **Bot-check**: some videos trigger YouTube's "Sign in to confirm you're not a bot" on fully anonymous requests. The documented commands had no cookie fallback.
2. **JS challenge / format resolution**: independent of bot-check, YouTube's JS challenge (needed to resolve playable video/audio formats) sometimes can't be solved locally without an extra remote component. When that happens, yt-dlp hard-fails with `Requested format is not available` — even for metadata/subtitle/comment reads that never needed a playable format to begin with.

Both are easy to hit on an arbitrary video and, before this fix, would cause the documented commands to fail outright with no guidance on why or what to do next.

## Fix

- Add `--ignore-no-formats-error` to every read-only yt-dlp invocation in the docs (metadata, subtitles, comments, search) — safe since none of these operations need an actual playable video/audio stream.
- Document a `--cookies-from-browser` retry as the fallback specifically for the bot-check case, scoped to local/desktop environments (it doesn't apply on headless servers, which get pointed at the Whisper-based fallback instead).
- Clarify that once past both issues, "no subtitles for the requested languages" often just means the video genuinely has none — pointing at `--list-subs` to confirm before assuming it's a config problem, and `agent-reach transcribe` as the audio-based fallback.

Updated: `agent_reach/skill/SKILL.md` + `SKILL_en.md` (quick-command pointers), `agent_reach/skill/references/video.md` (full fallback flow), `docs/troubleshooting.md` (human-facing entry, matching the existing Xueqiu/Twitter troubleshooting format).

## Test plan

- [x] `pytest tests/` — 196 passed, no regressions
- [x] Verified against a real video that was hitting both failure modes end-to-end:
  - `yt-dlp --ignore-no-formats-error --dump-json URL` — now succeeds (bot-check downgrades to a non-fatal warning, full metadata returned) where it previously hard-failed
  - `yt-dlp --ignore-no-formats-error --cookies-from-browser chrome --write-sub ... URL` — completes and correctly reports "no subtitles for the requested languages" (this particular video has none in any language) instead of crashing with the format error

### DIFF
--- a/agent_reach/skill/SKILL.md
+++ b/agent_reach/skill/SKILL.md
@@ -80,8 +80,8 @@ curl -s "https://r.jina.ai/URL"
 # GitHub 搜索
 gh search repos "query" --sort stars --limit 10
 
-# YouTube 字幕（注意：B站不要用 yt-dlp，见 video.md）
-yt-dlp --write-sub --skip-download -o "/tmp/%(id)s" "URL"
+# YouTube 字幕（注意：B站不要用 yt-dlp，见 video.md；被风控要求登录时的 cookie 兜底也在 video.md）
+yt-dlp --ignore-no-formats-error --write-sub --skip-download -o "/tmp/%(id)s" "URL"
 
 # V2EX 热门
 curl -s "https://www.v2ex.com/api/topics/hot.json" -H "User-Agent: agent-reach/1.0"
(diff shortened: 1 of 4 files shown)


Panniantong:
感谢分析 YouTube bot-check 和 JS challenge。当前 main 已升级 yt-dlp、加入版本门禁与可执行诊断；本 PR 的大范围 challenge workaround 已过期且缺少当前真实验收，因此关闭。

PR #460, opened 2026-07-01
fix(transcribe): retry with browser cookies on YouTube bot-check

HairtonicDublin:
## Summary

`agent-reach transcribe` (and `YouTubeChannel.transcribe()`) downloads audio via `yt-dlp -x` with no cookie handling at all, so any video that triggers YouTube's "Sign in to confirm you're not a bot" check on anonymous requests failed outright — independent of whether a Groq/OpenAI key was configured.

Separately, `configure youtube-cookies` already stores a browser preference in the `youtube_cookies_from` config key, but nothing in the codebase ever reads it — it's a write-only setting. It prints a success message and looks like it worked, but has zero effect on yt-dlp's actual behavior. (Grepped the whole codebase to confirm — only the `configure` handler sets it, nothing consumes it.)

## Fix

- `download_audio()` now retries once with `--cookies-from-browser` when the first attempt fails specifically with YouTube's bot-check message, using the configured `youtube_cookies_from` browser (defaulting to `"chrome"` if unset) — finally wiring up that config key.
- Detection is based on the full, untruncated stderr (attached to the raised `TranscribeError` as `.stderr`) rather than the display message, which is capped at 300 chars and could otherwise clip the match on longer warning output preceding the bot-check line.
- This only helps in environments with a local logged-in browser; on a headless server the retry itself fails and the original bot-check error surfaces, same as before.
- Updated the `configure youtube-cookies` success message to mention the bot-check use case alongside the existing age-restricted/member-video one.

## Test plan

- [x] `pytest tests/` — 200 passed (196 existing + 4 new)
- [x] New `TestDownloadAudioBotCheckRetry` class covers: retry with a configured browser, default-to-`chrome` fallback, no retry on unrelated failures, and the no-config-passed case
- [x] Updated two existing test doubles (`test_auto_temp_dir_is_cleaned_up`, `test_explicit_out_dir_is_preserved`) whose fake `download_audio()` needed the new `config` parameter in its signature
- [x] `ruff check` on all touched files — no new issues (pre-existing unrelated lint debt elsewhere in `cli.py` untouched)
- [x] Verified end-to-end against a real video that was hitting this exact failure: `agent-reach transcribe` now succeeds and produces a full transcript via Groq Whisper, where it previously failed immediately on the bot-check error

Note: getting an actual playable audio stream for that same test video also required yt-dlp's `--remote-components ejs:github` (their own official JS-challenge solver) — a separate, unrelated YouTube-side restriction on some videos that this fix doesn't touch or control.

### DIFF
--- a/agent_reach/cli.py
+++ b/agent_reach/cli.py
@@ -1114,7 +1114,8 @@ def _cmd_configure(args):
     elif args.key == "youtube-cookies":
         config.set("youtube_cookies_from", value)
         print(f"✅ YouTube cookie source configured: {value}")
-        print("   yt-dlp will use cookies from this browser for age-restricted/member videos.")
+        print("   yt-dlp will use cookies from this browser for age-restricted/member videos,")
+        print("   and as the automatic retry when a video triggers YouTube's bot-check.")
 
     elif args.key == "xhs-cookies":
         _configure_xhs_cookies(value)
--- a/agent_reach/transcribe.py
+++ b/agent_reach/transcribe.py
@@ -77,9 +77,11 @@ def _run(cmd: List[str], timeout: int = 600) -> None:
     except subprocess.TimeoutExpired:
         raise TranscribeError(f"{cmd[0]} timed out after {timeout}s")
     if proc.returncode != 0:
-        raise TranscribeError(
+        err = TranscribeError(
             f"{cmd[0]} failed (exit {proc.returncode}): {proc.stderr.strip()[:300]}"
         )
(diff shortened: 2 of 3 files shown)


Panniantong:
感谢处理 YouTube bot-check。自动读取并重试浏览器 Cookie 会扩大凭据边界，当前策略要求用户显式管理认证；merged #575 和 #577 已提供真实诊断与安全转录边界，因此关闭该实现。

PR #465, opened 2026-07-02
fix(opencli): diagnose OPENCLI_DAEMON_PORT conflict instead of blaming node

suyash-lyzr:
## Problem

On machines with the **OpenCLIApp** desktop build, every OpenCLI-backed channel (Reddit / Facebook / Instagram / XiaoHongShu / Bilibili) is reported as broken with a misleading fix:

> `opencli 命令存在但无法执行（node 环境损坏），重装：npm install -g @jackwener/opencli`

The node environment is **fine**. The real cause: `opencli >=1.8` hard-exits with `EX_CONFIG` (78) the moment `OPENCLI_DAEMON_PORT` is set — *even to the correct default port* — and the guard runs *before* `--version`, so every probe (including `--version`) fails. The OpenCLIApp desktop build still injects `OPENCLI_DAEMON_PORT=19825` into the CLI's environment, so a fresh macOS app install is effectively dead-on-arrival, and `npm install` does nothing to fix it.

## Fix

Detect the rejection from the probe output and prescribe the actual remedy (`unset OPENCLI_DAEMON_PORT`), noting node is fine — instead of a useless reinstall.

- Adds `env_conflict` to `OpenCLIStatus` + a dedicated `opencli_summary` line
- Correct, actionable hint (`unset OPENCLI_DAEMON_PORT`) rather than "reinstall node"
- Regression test; full suite green (197 passed)

## Verification

Reproduced live against the real OpenCLIApp shim: exit code `78`, output contains `OPENCLI_DAEMON_PORT is no longer supported (received 19825)`. The patched probe now returns `env_conflict=True` and the summary/hint direct the user to `unset OPENCLI_DAEMON_PORT`.

### DIFF
--- a/agent_reach/backends/opencli.py
+++ b/agent_reach/backends/opencli.py
@@ -28,6 +28,19 @@ OPENCLI_EXTENSION_URL = (
     f"https://chromewebstore.google.com/detail/opencli/{OPENCLI_EXTENSION_ID}"
 )
 
+#: Deprecated env var. Newer opencli (>=1.8) hard-exits (EX_CONFIG / 78) the
+#: moment OPENCLI_DAEMON_PORT is set — even to the correct default port — and
+#: the guard runs before `--version`, so every probe fails. The OpenCLIApp
+#: desktop build still injects it, so a fresh macOS app install is DOA until
+#: the var is unset. Detect this so we prescribe `unset` instead of a useless
+#: "reinstall node" — nothing is actually broken.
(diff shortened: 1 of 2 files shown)


Panniantong:
感谢定位 OPENCLI_DAEMON_PORT 环境冲突。main 已在子进程探测中剥离该变量并保留宿主环境，回归测试已覆盖，因此关闭。

PR #493, opened 2026-07-11
fix(installer): use platform-specific yt-dlp config path

yupoet:
## What changed

- Reuse `get_ytdlp_config_path()` when the installer configures the Node.js runtime for yt-dlp.
- Add a regression test that simulates macOS and verifies the installer writes `--js-runtimes node` under `~/Library/Application Support/yt-dlp/config`.

## Why

In v1.5.0, `_install_system_deps()` hardcodes the Linux path `~/.config/yt-dlp/config`, while the YouTube doctor check uses the platform-aware path helper. On macOS this makes installation report success, but the following `agent-reach doctor` cannot find the setting and warns that the JS runtime is not configured.

Using the existing shared helper keeps installation and health checks aligned on every supported platform.

## User impact

After `agent-reach install --env=auto` on macOS, YouTube readiness is detected correctly without requiring users to duplicate the yt-dlp setting manually.

## Validation

- `python3 -m pytest tests/ -q` — 197 passed
- `python3 -m ruff check --select F401 agent_reach/cli.py tests/test_cli.py` — passed

### DIFF
--- a/agent_reach/cli.py
+++ b/agent_reach/cli.py
@@ -9,13 +9,14 @@ Usage:
     agent-reach setup
 """
 
-import sys
 import argparse
 import json
 import os
+import sys
 import time
 
 from agent_reach import __version__
+from agent_reach.utils.paths import get_ytdlp_config_path
 
 # Pinned to the 0.4.2 state — PyPI still only has 0.4.1 (upstream issue #10).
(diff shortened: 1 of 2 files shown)


Panniantong:
感谢指出 yt-dlp 配置路径的跨平台问题。merged #575 已统一 HOME-first resolver 并覆盖 Windows 和 WSL 路径，因此关闭。

PR #495, opened 2026-07-11
fix(paths): emit single-line yt-dlp JS-runtime fix command

nankingjing:
### Problem

`render_ytdlp_fix_command()` returns the shell command that `agent-reach doctor`
tells users to run when yt-dlp is installed with Node.js but the
`--js-runtimes` config is missing (see `YouTubeChannel.check()` — a common,
documented case).

The POSIX branch builds the command with an f-string:

```python
f"printf '%s\n' '--js-runtimes node' >> '{config_path}'"
```

Here `\n` is a Python source escape, so it becomes an **actual newline
character** inside the returned string. The "command" is therefore split
across two lines, with line 1 ending in an unclosed single quote:

```
... 2>/dev/null || printf '%s
' '--js-runtimes node' >> '.../config'
```

In the doctor/warn output this renders as a broken two-line fragment. A user
who copies the visible first line pastes `... || printf '%s` with a dangling
quote and the shell hangs waiting for the quote to close.

### Root cause

The intent was the classic `printf '%s\n'` idiom, where `\n` is the **shell**
escape (a literal backslash-n), not a Python newline. The backslash was not
escaped in the f-string.

### Fix

One character — `%s\n` → `%s\n` — so the emitted command contains the literal
`\n` shell escape and stays on a single line:

```
... 2>/dev/null || printf '%s\n' '--js-runtimes node' >> '.../config'
```

Runtime behavior when executed is unchanged: `printf '%s\n'` still appends the
single line `--js-runtimes node`, and the `grep -qxF` guard keeps it
idempotent. The Windows (PowerShell) branch is untouched.

### Verification

- `python -m py_compile agent_reach/utils/paths.py` — OK.
- Added `tests/test_paths.py`; `pytest` — 2 passed. The single-line regression
  test **fails** on the current code and **passes** with this fix.


nankingjing:
Ready for review. One-character fix: `%s\n` → `%s\n` in the yt-dlp fix command so the emitted command stays on a single line instead of breaking at a literal newline (which causes a dangling shell quote). Two regression tests cover the single-line assertion. @moxxxskaifa PTAL.

Panniantong:
感谢发现 yt-dlp 修复命令中的换行问题。main 已用更强实现覆盖：输出保持单行、路径经过 shell quoting、逻辑正确分组，并有真实执行两次的幂等测试，因此关闭此已取代 PR。

PR #497, opened 2026-07-12
docs(youtube): add OpenCLI transcript fallback

sitabanubanu:
## Summary

Document an OpenCLI Browser Bridge fallback for YouTube transcript extraction when
the default `yt-dlp` route hits a bot-check, JavaScript challenge, or an
anonymous subtitle failure.

This keeps `yt-dlp` as the default route and adds a documented browser-session
fallback:

```bash
opencli doctor
opencli youtube transcript "https://www.youtube.com/watch?v=VIDEO_ID" -f json
```

The returned JSON contains timestamped transcript segments that an agent can
pass to downstream analysis.

## Changes

- Add the fallback command to both localized skill entry points.
- Add the full route, limitations, and troubleshooting guidance to the video
  reference and troubleshooting guide.
- Add a skill-resource regression assertion so both locales keep the route.
- Make the existing Windows YouTube config-path test deterministic by mocking
  the current platform-aware helper instead of the deprecated `expanduser()`
  path.

## Real-world validation

On a Windows desktop with OpenCLI v1.8.6 and a connected Browser Bridge:

- `opencli doctor` reported the daemon and extension connected.
- `opencli youtube transcript "https://www.youtube.com/watch?v=eWFKPPgHMCw" -f json`
  succeeded and returned a timestamped transcript for the full video.
- The downstream source-gated workflow then admitted the transcript as primary
  video material; its browser-first decision is documented in the companion
  project: https://github.com/sitabanubanu/codex-knowledge-workflow-skills/blob/codex/acquisition-bundle-v2/docs/adr/0003-browser-bridge-primary-acquisition.md

## Tests

- `pytest tests/ -q` -> **185 passed, 11 skipped**
- `pytest tests/test_skill_command.py -q` -> **5 passed**
- `ruff check tests/test_channel_contracts.py tests/test_skill_command.py` -> passed
- `python -m py_compile agent_reach/cli.py` -> passed
- `git diff --check` -> passed

## Boundary

This only uses a browser session the user has already authorized. It does not
bypass CAPTCHA, paywalls, private access, account permissions, or regional
restrictions. If the browser session has no usable transcript, the existing
Whisper fallback remains the next route.

This complements the current bot-check and cookie-handling work in #459 and
#460; it does not replace `yt-dlp` or add a new report-generation layer.

### DIFF
--- a/agent_reach/skill/SKILL.md
+++ b/agent_reach/skill/SKILL.md
@@ -83,6 +83,9 @@ gh search repos "query" --sort stars --limit 10
 # YouTube 字幕（注意：B站不要用 yt-dlp，见 video.md）
 yt-dlp --write-sub --skip-download -o "/tmp/%(id)s" "URL"
 
+# YouTube 浏览器会话字幕兜底（yt-dlp 遇到 bot-check / JS challenge 时）
+opencli youtube transcript "URL" -f json
+
 # V2EX 热门
 curl -s "https://www.v2ex.com/api/topics/hot.json" -H "User-Agent: agent-reach/1.0"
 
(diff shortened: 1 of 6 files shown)


Panniantong:
感谢补充 YouTube 转录 fallback 文档。main 已按当前 OpenCLI 与 yt-dlp 能力重写真实状态和降级说明，本旧文档分支已过期，因此关闭。

PR #517, opened 2026-07-19
fix: bump dead yt-dlp pin and gate doctor's --js-runtimes prescription on version support

### LINKED ISSUE #503: v1.5.0: doctor prescribes --js-runtimes but constraints.txt pins yt-dlp==2025.5.22 (flag unsupported; pin also fails 2026 YouTube)
> **TL;DR（中文）**：v1.5.0 的 doctor 建议写入 `--js-runtimes node`，但仓库自己的 `constraints.txt` 钉死的 `yt-dlp==2025.5.22` 不支持该参数；而且这个旧版对当前 YouTube 已经无法解析。升级 yt-dlp 后一切正常，但再按 `-c constraints.txt` 重装会把修好的版本降级回坏的。

**TL;DR (EN):** In v1.5.0, `doctor`'s prescribed yt-dlp config flag (`--js-runtimes node`) is not supported by the repo's own pinned `yt-dlp==2025.5.22`; that pin itself also fails against live 2026 YouTube; and re-installing with `-c constraints.txt` downgrades a user-fixed setup back to broken.

## Environment

- agent-reach **v1.5.0** (installed from the `v1.5.0` tag), deps installed with `pip install -c constraints.txt ...` per the header in `constraints.txt`
- Windows 11 (native), Python 3.14
- Node.js v24 present, **no deno** (this is what triggers the Node prescription path)
- yt-dlp **2025.5.22** (the constraints pin)

## Steps to reproduce

1. Install from tag `v1.5.0` honoring the tested pin set → `yt-dlp==2025.5.22`.
2. Run `agent-reach doctor` (or `agent-reach install`) on a machine with Node but without deno → the YouTube channel warns and prescribes writing `--js-runtimes node` into the yt-dlp config (`%APPDATA%\yt-dlp\config` on Windows).
3. Apply the prescription exactly as printed → **every** yt-dlp invocation now fails:

   ```
   yt-dlp: error: no such option: --js-runtimes
   ```

4. Remove the config line → extraction still fails against live YouTube with the pinned version:

   ```
   ERROR: YoutubeIEContentProviderLogger.debug() got an unexpected keyword argument 'once'
   ```

5. `pip install -U yt-dlp` (tested with **2026.7.4**) + re-apply the same prescription → works. Verified end-to-end: video title and `.en.vtt` auto-subtitles download successfully.
6. Re-run the documented install command **with** `-c constraints.txt` → pip downgrades yt-dlp back to 2025.5.22 → broken again.

## Expected vs actual

- **Expected:** the pinned dependency set and doctor's prescriptions are mutually consistent; applying doctor's advice never breaks a channel.
- **Actual:** the prescription is rejected by the pinned yt-dlp (`no such option: --js-runtimes`), and the pin itself can no longer parse 2026 YouTube (`YoutubeIEContentProviderLogger.debug() ... 'once'`).

Minor related observation: `doctor` can show YouTube as ✅ while extraction is actually broken, because the probe is `yt-dlp --version`, which succeeds even with an invalid config option or a stale extractor.

## Root cause (v1.5.0 tag sources)

- `constraints.txt:11` → `yt-dlp==2025.5.22`
- `agent_reach/channels/youtube.py:57-64` → when Node is the JS runtime, doctor warns and prints the fix command
- `agent_reach/utils/paths.py:29-46` (`render_ytdlp_fix_command`) → renders the `--js-runtimes node` config append

`--js-runtimes` only exists in newer yt-dlp releases (present in 2026.7.4; rejected by 2025.5.22), so v1.5.0 prescribes a flag its own tested pin cannot parse. Independently, yt-dlp is a rolling extractor — a ~14-month-old pin cannot track YouTube's changes.

## Suggested fix

1. Bump the `constraints.txt` yt-dlp pin to a current release (≥ 2026.x), and ideally add a CI smoke test that runs the pinned yt-dlp against a live YouTube URL so a dead pin is caught automatically; **or**
2. Gate the `--js-runtimes` prescription on the detected yt-dlp version (only prescribe when the flag is supported), and have doctor warn when the installed yt-dlp is significantly stale.
3. Docs: note that re-installing with `-c constraints.txt` will downgrade a user-upgraded yt-dlp (easy footgun after a manual fix).

Related: #378 mentions the stale yt-dlp pin in passing (security/staleness angle); this issue is the concrete functional break plus the self-contradicting doctor prescription.

mvanhorn:
## Summary

`doctor` on a Node-only machine prescribes appending `--js-runtimes node` to the yt-dlp config, but `constraints.txt` pins `yt-dlp==2025.5.22`, which doesn't support that flag, so every subsequent yt-dlp call dies with `error: no such option: --js-runtimes`. This bumps the pin to a current release and gates the prescription on the installed yt-dlp version so the tested pin and doctor's advice can no longer contradict each other.

## Background

The reporter (ayoubagrebi062-hue, #503) verified two coupled problems: the ~14-month-old pin can no longer parse 2026 YouTube (`YoutubeIEContentProviderLogger.debug() got an unexpected keyword argument 'once'`), and even after upgrading yt-dlp by hand, re-running the documented `pip install -c constraints.txt ...` downgrades it back to the broken pin. Meanwhile `YouTubeChannel.check()` prescribes `--js-runtimes`, a flag `2025.5.22` rejects.

This makes two coordinated changes, mapping to the reporter's own suggestions:

- `constraints.txt`: pin `yt-dlp==2026.7.4` (the reporter-verified working version), which parses 2026 YouTube and supports `--js-runtimes`.
- `YouTubeChannel.check()`: the probe already runs `yt-dlp --version`, so parse that in-process and only prescribe `--js-runtimes` when the installed version is at/above the supported floor. Below the floor it returns a `warn` telling the user to `pip install -U yt-dlp` instead of printing a flag the binary will reject; empty or unparseable version output falls back to today's prescription so nothing regresses. `paths.py` (which renders the fix command) is untouched.

Added tests to `tests/test_youtube_channel.py` covering the supported, stale, and unparseable-version paths, the unchanged deno path, and a consistency guard asserting `constraints.txt` pins a 2026-or-later release. `pytest tests/test_youtube_channel.py` passes (16 tests).

Closes #503

### DIFF
--- a/agent_reach/channels/youtube.py
+++ b/agent_reach/channels/youtube.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 """YouTube — check if yt-dlp is available with JS runtime."""
 
+import re
 import shutil
 
 from agent_reach.probe import probe_command
@@ -8,6 +9,8 @@ from agent_reach.utils.paths import get_ytdlp_config_path, render_ytdlp_fix_comm
 from agent_reach.utils.text import read_utf8_text
 
 from .base import Channel
+
+_JS_RUNTIMES_SUPPORTED_FROM = (2026, 7, 4)
 
 
 def _has_js_runtime_config(config_path) -> bool:
@@ -18,6 +21,12 @@ def _has_js_runtime_config(config_path) -> bool:
         return "--js-runtimes" in read_utf8_text(config_path)
     except OSError:
         return False
+
+
+def _parse_ytdlp_version(version: str):
+    """Return a comparable yt-dlp version tuple, or ``None`` if unknown."""
+    match = re.fullmatch(r"\s*(\d+)\.(\d+)\.(\d+)\s*", version)
+    return tuple(map(int, match.groups())) if match else None
 
 
 class YouTubeChannel(Channel):
@@ -60,6 +69,12 @@ class YouTubeChannel(Channel):
         if not has_deno:
             ytdlp_config = get_ytdlp_config_path()
             if not _has_js_runtime_config(ytdlp_config):
+                version = _parse_ytdlp_version(probe.output)
+                if version is not None and version < _JS_RUNTIMES_SUPPORTED_FROM:
+                    return "warn", (
+                        "yt-dlp 版本过旧，无法配置 JS runtime。请升级：\n"
+                        "  pip install -U yt-dlp"
+                    )
                 return "warn", (
                     f"yt-dlp 已安装但未配置 JS runtime。运行：\n  {render_ytdlp_fix_command()}"
                 )
@@ -88,4 +103,3 @@ class YouTubeChannel(Channel):
         from agent_reach.transcribe import transcribe as _transcribe
 
         return _transcribe(url, provider=provider, config=config)
-
--- a/constraints.txt
+++ b/constraints.txt
@@ -8,7 +8,7 @@ python-dotenv==1.2.1
 loguru==0.7.3
 PyYAML==6.0.3
 rich==14.3.2
-yt-dlp==2025.5.22
+yt-dlp==2026.7.4
 
 pytest==8.0.0
 ruff==0.15.1
--- a/tests/test_youtube_channel.py
+++ b/tests/test_youtube_channel.py
@@ -11,11 +11,12 @@ channel coverage after rss (#360), github (#361), web (#363),
 reddit (#364), xueqiu (#365) and v2ex (#366).
 """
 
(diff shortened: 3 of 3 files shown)


czwziy:
你好，来信已收到，谢谢！czwziy

Panniantong:
感谢准确指出 yt-dlp pin 与 --js-runtimes 的矛盾。main 已完成版本及运行时修复，因此关闭此已覆盖 PR。

PR #518, opened 2026-07-19
fix(doctor): detect OpenCLI extension on Edge and unpacked installs

### LINKED ISSUE #513: doctor: false negative for OpenCLI extension loaded as unpacked in Edge (not Chrome)
## Summary
`agent-reach doctor --json` reports the OpenCLI extension as "not installed" for **reddit / facebook / instagram / xiaohongshu**, even though the extension **is installed and connected**. In this setup the extension is loaded as an **unpacked extension in Microsoft Edge (developer mode)**, not installed from the Chrome Web Store into Chrome.

## Environment
- agent-reach v1.5.0 (Windows 11)
- OpenCLI v1.8.6, extension v1.0.22
- Extension loaded as **unpacked in Edge dev mode** (files at `~/.opencli/extension/`)
- Google Chrome is installed but has **no `User Data` directory** (never initialized)

## What doctor reports
For reddit / facebook / instagram / xiaohongshu:
```
status: warn
message: "OpenCLI 已安装，但 Chrome 扩展未安装。 1. 安装扩展（需手动点一次）：https://chromewebstore.google.com/... 2. 保持 Chrome 打开，运行 `opencli doctor` 验证"
```

## What's actually true
`opencli doctor` confirms everything is healthy:
```
[OK] Daemon: running on port 19825 (v1.8.6)
[OK] Extension: connected (v1.0.22)
Profiles:
  • 7ghw3ueb: connected v1.0.22
[OK] Connectivity: connected in 8.4s
Everything looks good!
```
The daemon on port 19825 also responds to requests, confirming the extension is live. These 4 platforms actually work.

## Likely root cause
`doctor` appears to detect the OpenCLI extension by scanning Chrome's standard `User Data\<Profile>\Extensions\` directory for the extension ID (`ildkmabpimmkaediidaifkhjpohdnifk`). This fails when:
1. The user runs **Edge** instead of Chrome (Chrome's `User Data` doesn't exist on this machine).
2. The extension is loaded as an **unpacked extension** from `~/.opencli/extension/` rather than installed from the Web Store.

## Suggested fix
Instead of (or in addition to) scanning the browser extension directory, query the **OpenCLI daemon** (port 19825, with the proper `X-OpenCLI` header) to determine whether the extension is connected — `opencli doctor` already does this correctly and could serve as the reference.

## Workaround
Ignore the `warn` for these 4 platforms; they work. Run `opencli doctor` for the real status.

## Related
- #451 (extension genuinely not connecting — opposite direction)
- #159 (closed: doctor falsely reported xiaohongshu MCP failure on Windows — same false-negative pattern)

mithunbabbira:
## Summary
- Widen OpenCLI on-disk extension detection to Microsoft Edge profile roots (macOS/Linux/Windows).
- Treat `~/.opencli/extension/manifest.json` as an unpacked install so sleeping extensions no longer false-negative in `doctor`.
- Soften the missing-extension hint so it is not Chrome-only.

Fixes #513.

## Test plan
- [x] `pytest tests/test_opencli_backend.py -v`
- [x] `pytest tests/ -q`
- [ ] Manual (optional): Edge + Load unpacked from `~/.opencli/extension`, leave extension disconnected/sleeping, confirm `agent-reach doctor --json` reports reddit/facebook/instagram/xiaohongshu ready when disk path exists

Made with [Cursor](https://cursor.com)

### DIFF
--- a/agent_reach/backends/opencli.py
+++ b/agent_reach/backends/opencli.py
@@ -13,7 +13,8 @@ Probing notes (verified live):
     service worker sleeps and any real opencli command wakes it up
     (verified: status flips disconnected→connected after one call).
     Since daemon status can't tell "sleeping" from "never installed",
-    we check Chrome's Extensions directory on disk to disambiguate.
+    we check Chrome/Edge/Chromium Extensions directories and OpenCLI's
+    unpacked layout on disk to disambiguate.
 """
 
 import glob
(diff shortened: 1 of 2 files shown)


Panniantong:
感谢补充 Edge 与 unpacked OpenCLI 场景。main 已覆盖对应 profile 和 unpacked manifest，同时保留磁盘文件不等于实时可用的安全语义，因此关闭。

PR #526, opened 2026-07-23
fix(cookie-extract): let --from-browser target a specific Chrome profile

LucasRollo:
## Problem

`agent-reach configure --from-browser chrome` (and the auto-import during install) silently reads cookies from the **Default** Chrome profile only, with no way to pick another profile.

Both extraction backends resolve to the first matching cookie DB and stop:
- `browser_cookie3`'s Chrome path list is `['.../Default/Cookies', '.../Profile */Cookies']` and it returns the first glob match (`next(...)`), so `Default` always wins when present.
- `rookiepy` does the same in its Rust `find_chrome_based_paths` (first existing path wins, early `return`), and exposes no profile parameter through its Python API.

This directly undercuts the project's own security guidance. The README and `docs/install.md` recommend using a **dedicated/secondary account** for cookie-auth platforms (Twitter, XHS, Reddit, etc.). The natural way to do that is a separate Chrome profile for burner accounts — but with this bug, extraction ignores that profile and grabs cookies from the user's real **Default** profile instead. The safe workflow the docs recommend can't actually be followed.

## Fix

- **`list_chrome_profiles(browser)`** — enumerates profiles from the browser's own `Local State` file (the same source Chrome's profile picker uses), returning each profile's folder, display name, account email, and exact `Cookies` DB path. Supports chrome/edge/brave/opera across macOS/Windows/Linux.
- **`extract_all()` / `configure_from_browser()`** — new optional `profile=` parameter. When set, it resolves that profile's exact cookie DB and routes through `browser_cookie3(cookie_file=...)` (rookiepy has no profile override). When omitted, behavior is **byte-for-byte unchanged** — no regression for existing users.
- **CLI `--chrome-profile` flag** — when multiple profiles exist and none is specified: prompts interactively (numbered list with name + email) at a TTY; otherwise lists the available folders and aborts rather than silently guessing.

## Tests

- `tests/test_cookie_extract_profiles.py` — profile discovery, missing/empty `Local State`, profiles without a `Cookies` DB yet, the core regression (right profile → right cookie path), and the two error paths (unknown profile, firefox+profile).
- Full existing suite still passes (202 passed on Python 3.12).

## Example

```
$ agent-reach configure --from-browser chrome
Found 2 chrome profiles:
  1. Lucas (main) (real@gmail.com)  [Default]
  2. Research (burner@gmail.com)  [Profile 1]
Which profile has the accounts you want? [1-2]: 2

# or non-interactively:
$ agent-reach configure --from-browser chrome --chrome-profile "Profile 1"
```

### DIFF
--- a/agent_reach/cli.py
+++ b/agent_reach/cli.py
@@ -89,11 +89,12 @@ def main():
                         choices=["chrome", "firefox", "edge", "brave", "opera"],
                         help="Auto-extract ALL platform cookies from browser (chrome/firefox/edge/brave/opera)")
     p_conf.add_argument("--chrome-profile", metavar="PROFILE", default=None,
-                        help="Profile folder to extract from with --from-browser on a "
-                             "Chromium-based browser (e.g. 'Profile 1'), for when you keep "
+                        help="Profile folder to extract from with --from-browser on "
+                             "chrome/edge/brave (e.g. 'Profile 1'), for when you keep "
                              "burner/research accounts in a separate profile from your main "
-                             "one. Without this, extraction silently reads the 'Default' "
-                             "profile. Omit to be prompted when more than one profile exists.")
+                             "one. Without this, extraction defaults to the 'Default' profile; "
+                             "if several profiles exist you'll be prompted to pick one when "
+                             "running interactively, or asked to pass this flag otherwise.")
 
     # ── doctor ──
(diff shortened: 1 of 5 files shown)


Panniantong:
感谢指出多 Chrome profile 场景。merged #530 已提供显式 --profile，且找不到目标时 fail closed，因此关闭此已覆盖 PR。

PR #527, opened 2026-07-23
test: stop the suite from writing skill files into the real $HOME

LucasRollo:
## Problem

Running the test suite mutates the developer's real home directory.

`tests/test_cli.py::test_doctor_runs` invokes the real `agent-reach doctor`:

```python
def test_doctor_runs(self, capsys):
    with patch("sys.argv", ["agent-reach", "doctor"]):
        main()
    ...
```

`_cmd_doctor` calls `_install_skill(force=False)` ("Auto-install skill if not already present"), which installs `SKILL.md` into every agent directory it finds on the machine — `~/.claude/skills/agent-reach/`, `~/.agents/skills/agent-reach/`, `~/.openclaw/skills/agent-reach/` — and `Config()` creates `~/.agent-reach/`. The test patches only `sys.argv`, so all of that lands in the developer's **real** `$HOME`.

That means simply running `pytest` silently installs (or overwrites) skill files for the developer's live AI agents. A SKILL.md is standing instructions for an agent, so this isn't just stray files — it can change how the developer's Claude Code / OpenClaw behaves afterward. It's a surprising side effect, and the same pattern a malicious dependency could use to inject agent instructions via a test run.

The sibling test `test_doctor_preserves_existing_skill_install` already carefully sandboxes `expanduser` and the `Config` paths, so the need for isolation is understood — `test_doctor_runs` just lacked it, and nothing enforced it suite-wide.

## Fix

Rather than patch a single test, isolate `$HOME` for the whole suite so no test can leak into the real home:

- **`tests/conftest.py`** — autouse fixture redirecting `HOME`/`USERPROFILE` to a throwaway tmp dir. It also re-points `Config.CONFIG_DIR` / `Config.CONFIG_FILE`, which are class attributes bound to `Path.home()` at **import time** (before any fixture runs), so an env-var redirect alone wouldn't contain them.
- **`tests/test_home_isolation.py`** — asserts the isolation is actually in effect (expanduser + Config resolve under the sandbox; doctor's writes stay under it). If the fixture is ever removed or stops redirecting `$HOME`, these fail instead of silently leaking again.

## Verification

Full suite passes, and after running it the real `~/.claude/skills/agent-reach` and `~/.agent-reach` are **no longer created** (previously they were).

## Note (out of scope for this PR)

Arguably a diagnostic command (`doctor`) shouldn't install skills into the system at all as a side effect — that's a surprising mutation from a read-only-sounding command. Happy to follow up separately if the maintainers want to reconsider that behavior; this PR just stops the test suite from being affected by it.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,35 @@
+# -*- coding: utf-8 -*-
+"""Shared pytest fixtures.
+
+Sandboxes $HOME for every test. Several commands write into the user's
+home directory as a side effect — most notably `agent-reach doctor`, which
+auto-installs SKILL.md into every agent directory it finds
+(~/.claude/skills, ~/.agents/skills, ~/.openclaw/skills) and creates
+~/.agent-reach/ via Config(). Without this fixture, simply running the
+test suite mutates the developer's real home directory and can silently
+install/overwrite skill files for their live AI agents.
+
+`os.path.expanduser("~")` and `pathlib.Path.home()` resolve from HOME on
+POSIX and USERPROFILE on Windows, so redirecting those env vars contains
+every *runtime* home-directed write (e.g. the skill-dir installs) under a
+throwaway tmp dir. Config.CONFIG_DIR / CONFIG_FILE are the exception: they
+are class attributes bound to Path.home() at import time — before any
+fixture runs — so they must be re-pointed explicitly.
+"""
+
+import pytest
+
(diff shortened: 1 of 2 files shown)


Panniantong:
感谢发现测试会写入真实 HOME。main 已完成 pytest HOME 隔离，因此关闭此已覆盖 PR。

PR #536, opened 2026-07-27
fix(v2ex): recover from Python TLS EOF

### LINKED ISSUE #514: doctor: false negative for v2ex (SSL EOF error) while API is reachable via curl
## Summary
`agent-reach doctor --json` reports v2ex as unreachable with an SSL error, but the V2EX API is fully reachable via `curl`. The SSL failure is in agent-reach's bundled Python `urllib`, not in the network.

## Environment
- agent-reach v1.5.0 (Windows 11)
- China network, no proxy

## What doctor reports
```
status: warn
name: "V2EX 节点、主题与回复"
message: "V2EX API 连接失败（可能需要代理）：<urlopen error [SSL: UNEXPECTED_EOF_WHILE_READING] EOF occurred in violation of protocol (_ssl.c:1010)>"
active_backend: null
```

## What's actually true
`curl` reaches the same endpoint successfully (HTTP 200 in ~2.5s, no proxy):
```
$ curl -sS -m 8 -o /dev/null -w "http_code=%{http_code} time=%{time_total}s\n" \
    "https://www.v2ex.com/api/topics/hot.json" -H "User-Agent: agent-reach/1.0"
http_code=200 time=2.519745s
```

## Likely root cause
The check uses Python's `urllib` (the `_ssl.c` reference in the error confirms this), which hits `SSL: UNEXPECTED_EOF_WHILE_READING`. `curl` uses a different TLS stack and succeeds. This is a false negative from the Python SSL environment bundled in the agent-reach executable, not a real connectivity problem. This also contributes to the ~32s `doctor` runtime (the failing check times out slowly).

## Suggested fix
- Use a more robust HTTP client / TLS config for the connectivity check (e.g. `requests` with proper cert handling, or shell out to `curl`).
- Or catch `SSL: UNEXPECTED_EOF_WHILE_READING` specifically and retry / fall back instead of reporting the channel as down.

## Workaround
Ignore the `warn` — v2ex is reachable. Use `curl` directly for V2EX API calls.

## Related
- #262 (closed: doctor false negative on Windows Chinese system for WeChat — same Windows + China-network false-negative pattern)

LeonChaoX:
## Summary

- cap V2EX `urllib` responses at 1 MiB
- retry only Python TLS unexpected-EOF failures through the native `curl` TLS stack
- preserve fail-closed behavior for certificate verification and ordinary network failures
- add regressions for fallback behavior and `doctor` recovery

## Root cause

On some Windows/network combinations, Python's TLS stack raises `UNEXPECTED_EOF_WHILE_READING` even though the same V2EX endpoint succeeds through the operating system's `curl`. Because the channel and doctor shared the same `urllib` path, Agent Reach reported V2EX as unavailable and could not fetch data.

## User impact

V2EX checks and data calls now recover when the native TLS stack can reach the endpoint, without weakening certificate validation or hiding unrelated failures.

## Validation

- `python -m pytest tests/test_v2ex_channel.py tests/test_doctor.py tests/test_channel_contracts.py tests/test_skill_command.py -q` — 37 passed
- `python -m ruff check agent_reach/channels/v2ex.py tests/test_v2ex_channel.py tests/test_skill_command.py`
- `git diff --check`

Closes #514

### DIFF
--- a/agent_reach/channels/v2ex.py
+++ b/agent_reach/channels/v2ex.py
@@ -2,22 +2,108 @@
 """V2EX — public API channel for topics, nodes, users, and replies."""
 
 import json
+import shutil
+import ssl
+import subprocess
 import urllib.request
 from typing import Any
 
+from agent_reach.utils.process import utf8_subprocess_env
 from agent_reach.utils.text import scrub_url_credentials
 
 from .base import Channel
 
 _UA = "agent-reach/1.0"
 _TIMEOUT = 10
+_MAX_RESPONSE_BYTES = 1024 * 1024
+
+
+def _get_json_with_urllib(url: str) -> Any:
+    """Fetch JSON with Python's standard HTTP stack."""
+    req = urllib.request.Request(url, headers={"User-Agent": _UA})
+    with urllib.request.urlopen(req, timeout=_TIMEOUT) as resp:
+        raw = resp.read(_MAX_RESPONSE_BYTES + 1)
+    if len(raw) > _MAX_RESPONSE_BYTES:
+        raise ValueError("V2EX API response exceeds the 1 MiB safety limit")
+    return json.loads(raw.decode("utf-8"))
+
+
+def _is_unexpected_tls_eof(error: BaseException) -> bool:
+    """Return whether an exception chain contains the retryable TLS EOF."""
+    pending: list[BaseException] = [error]
+    seen: set[int] = set()
+    while pending:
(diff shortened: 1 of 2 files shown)


Panniantong:
感谢实现 V2EX TLS EOF fallback。你的原始作者提交已保留在 merged #577 中，并在其上增加了精确错误边界、响应限制、回归测试和真实 smoke；因此关闭原 PR。

PR #542, opened 2026-07-29
test: make suite portable to Windows

hellojiaru:
## Summary

- add one Windows/Python 3.12 job for the full pytest suite
- make symlink security tests skip only when the host cannot create symlinks
- make the POSIX command test discover `sh` instead of assuming `/bin/sh`
- make home-directory and path-output assertions portable to Windows

## Problem

I ran the documented test workflow on Windows 11 at
`b4d52c46c9113cb0f653d6df4cf71ebadf4930ac`.

The unmodified suite completed with:

```text
9 failed, 393 passed, 26 skipped
```

The nine failures were test-environment assumptions rather than one runtime
bug:

- four tests attempted symlink creation on a host without symlink privilege;
- one POSIX behavior test invoked `/bin/sh` unconditionally;
- three tests overrode `HOME` but not Windows `USERPROFILE`;
- one assertion compared native and mixed-separator spellings of the same path.

The current pytest workflow only runs Ubuntu, so this class of regression is
not visible before merge.

## What changed

The patch keeps the security assertions intact:

- when symlink creation is available, the tests still exercise rejection of
  symlinked credential paths;
- when the host cannot create symlinks, pytest records an explicit capability
  skip instead of failing during setup;
- the POSIX command remains executed when a compatible `sh` is available.

The Windows job is intentionally limited to Python 3.12 to bound CI cost while
the existing Ubuntu job retains the Python 3.10–3.13 matrix.

## Validation

Windows 11 build 26200, Python 3.12.3:

```text
focused replay: 4 passed, 5 skipped
full suite:      397 passed, 31 skipped
changed files:   ruff check passed
```

The total remains 428 collected tests. `git diff --check` also passes.

For transparency, the repository-wide Ruff command still reports 20 existing
findings in unrelated files; this patch does not claim to fix that baseline.

## Boundaries

- no runtime product code changed
- no live platform or credential checks were used
- no unrelated formatting or lint cleanup is included

## AI assistance

Codex helped structure the Windows failure analysis and prepare the patch. I
ran every command reported above locally, reviewed the diff, and take
responsibility for the proposed change and follow-up.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -29,6 +29,26 @@ jobs:
         run: |
           pytest -q
 
+  windows-test:
+    runs-on: windows-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+
+      - name: Install package and test deps
+        run: |
+          python -m pip install --upgrade pip
+          pip install -c constraints.txt -e .[dev]
+
+      - name: Run tests
+        run: |
+          pytest -q
+
   # Editable installs (-e) never exercise wheel packaging, so a broken wheel
   # can pass tests and still fail every real `pip install` from source.
   # This job builds the actual wheel and installs it into a clean venv.
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -6,6 +6,20 @@ import pytest
 from agent_reach.config import Config
 
 
+@pytest.fixture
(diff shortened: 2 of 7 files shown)


Panniantong:
感谢贡献。Windows CI 的有效部分已在 #578 以保留你作者身份的提交 e0fc5b7 合并到 main；它实际发现并推动修复了 Git Bash 选择、Windows 路径、权限位和 UTF-8 解码等可移植性问题。最终两套 Windows 真机检查均通过，因此关闭此旧 PR。

PR #549, opened 2026-07-31
fix: decode yt-dlp and ffmpeg output as UTF-8

lukiod:
`yt-dlp` and `ffmpeg` emit UTF-8 on every platform, but `subprocess.run(..., text=True)` without an explicit `encoding` decodes with the locale codec, which is `gbk` on a Chinese Windows system.

A CJK video title then fails to decode inside subprocess's reader thread. `run` returns with `stdout`/`stderr` as `None`, and the caller raises `AttributeError: 'NoneType' object has no attribute 'strip'` instead of `TranscribeError`, after an unrelated thread traceback. The real error is lost.

Repro with a stand-in for yt-dlp that exits nonzero and writes a UTF-8 CJK title:

```
UnicodeDecodeError: 'gbk' codec can't decode byte 0xad in position 2: illegal multibyte sequence
RESULT: AttributeError: 'NoneType' object has no attribute 'strip'
```

Same root cause as #262, which fixed the `mcporter` call in `channels/wechat.py`. Two sites weren't covered: `_probe_audio_duration` (ffprobe) and `_run` (yt-dlp and ffmpeg, three callers).

Both now pass `encoding="utf-8", errors="replace"`, so an odd byte can't turn a subprocess failure into a crash that hides the original error.

Three tests decode CJK the way the real call does; all three fail on main. `ruff`, `mypy` and `pytest` give the same results before and after on my machine (20 ruff findings, 1 mypy error in `cli.py`, 9 failures, all pre-existing and POSIX-only). Tests go 393 → 396 passed.

Windows 11, Python 3.14.

### DIFF
--- a/agent_reach/transcribe.py
+++ b/agent_reach/transcribe.py
@@ -100,6 +100,8 @@ def _probe_audio_duration(path: Path) -> float:
             cmd,
             capture_output=True,
             text=True,
+            encoding="utf-8",
+            errors="replace",
             timeout=FFPROBE_TIMEOUT_SECONDS,
         )
     except subprocess.TimeoutExpired:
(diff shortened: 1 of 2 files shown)


Panniantong:
感谢贡献。该修复已在 #578 以保留你作者身份的提交 68a0f44 重实现并合并到 main；同时补了 GBK 默认环境行为测试和真实 Windows CI，两套 Windows 检查均已通过。因此关闭此旧 PR，贡献归属已保留。

PR #554, opened 2026-07-31
Report a failed container restart in xhs-cookies

lukiod:
`_configure_xhs_cookies` restarts the container so it reloads the cookies, but never checks the result:

```python
subprocess.run([docker, "restart", container_name], capture_output=True, encoding="utf-8", timeout=30)
print("done")
```

`subprocess.run` only raises on timeout or OSError, not on a non-zero exit, so `docker restart` failing (container gone, daemon down, permissions) still prints `done`. The `docker cp` right above it does check `returncode`.

The cookies are on disk by then, but xiaohongshu-mcp only reads them on restart. So the user is told configuration succeeded and keeps hitting auth failures with nothing pointing back here.

Now prints the stderr and the manual command, same shape as the `docker cp` failure path.

Two tests, one for each branch. The failure one fails on main. Suite goes 393 → 395 passed; `ruff` and `mypy` unchanged (20 findings and 1 pre-existing error in `cli.py`).

### DIFF
--- a/agent_reach/cli.py
+++ b/agent_reach/cli.py
@@ -1553,11 +1553,19 @@ def _configure_xhs_cookies(value) -> bool:
         # Restart container so it reloads cookies from disk
         print("  Restarting container to reload cookies...", end=" ", flush=True)
         try:
-            subprocess.run(
+            # subprocess.run does not raise on a non-zero exit, so the result has
+            # to be checked; the cookies are on disk but unread until the restart
+            # actually happens.
+            restart = subprocess.run(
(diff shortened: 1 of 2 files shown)


Panniantong:
感谢贡献。该修复已在 #578 以保留你作者身份的提交 7e9ff38 重实现并合并到 main；新版还会在重启失败时返回失败、输出受限错误详情，并保证临时 Cookie 文件清理。定向测试和全量 CI 已通过，因此关闭此旧 PR。

PR #555, opened 2026-07-31
fix: resolve symlinked $HOME so the config dir doesn't trip the symlink guard

initxy:
## Problem

On a host where `\$HOME` is itself a **symlink** — a relocated `/home` pointing at a larger disk, an NFS-mounted home, or a container bind-mount — **every** `agent-reach` command aborts immediately:

```
agent_reach.config.ConfigSecurityError: 配置目录不能经过符号链接：/home/<user>
```

### Reproduce

```bash
# /home/<user> -> /data00/home/<user>  (root-owned symlink, common on servers)
$ ls -ld /home/<user>
lrwxrwxrwx 1 root root 23 ... /home/<user> -> /data00/home/<user>

$ agent-reach doctor
ConfigSecurityError: 配置目录不能经过符号链接：/home/<user>
```

### Root cause

`Config.CONFIG_DIR` (and the per-backend credential paths in `github`/`reddit`/`xiaohongshu`/`mcporter`, plus `get_ytdlp_config_dir`) are built from `Path.home()`, which returns `\$HOME` **verbatim**. `Config.load()` then runs `ensure_no_symlink_path()`, which walks the path component-by-component with `os.lstat` and rejects it the moment *any* segment is a symlink. When `\$HOME` itself is the symlink, the check fires on the home segment and the process dies before doing anything.

## Fix

Add `utils.paths.user_home()`, which resolves **only the home prefix** via `os.path.realpath(Path.home())`, and route the six `Path.home()` call sites through it.

The symlink guard is **not weakened**. Everything agent-reach *creates* beneath home — `.agent-reach/`, `config.yaml`, cookie/credential files — is still checked component-by-component by `ensure_no_symlink_path`, so a symlink planted *inside* the home directory is still refused, and an explicitly-passed symlinked `config_path` is still rejected (existing tests `test_config_directory_symlink_is_rejected` / `test_config_ancestor_symlink_is_rejected` stay green).

### Why this is safe

The guard exists to stop a credential write/read from being **redirected** through an attacker-controlled symlink (TOCTOU). A system-supplied `\$HOME` symlink is a different thing: it is owned by root/the administrator, not the user. An attacker who could rewrite a root-owned `\$HOME` symlink already holds higher privileges than the agent-reach process — so resolving that one prefix yields **no** safety loss, while unblocking a large class of legitimate environments (relocated `/home`, NFS homes, containers).

Resolving the prefix also fixes a quieter bug: on a symlinked home, the YouTube JS-runtime detection (`get_ytdlp_config_dir` → `read_small_text_no_follow`) silently failed the same way.

## Tests

`tests/` (431 tests) pass. Added regression coverage:

- `test_user_home_resolves_symlinked_home` — `user_home()` resolves a symlinked home; a config path built beneath it is symlink-free end to end.
- `test_user_home_still_rejects_symlink_inside_home` — a link planted *inside* home is still refused (guard intact).
- `test_config_dir_under_symlinked_home_initializes_and_saves` — the original crash: `Config` now initializes **and** persists under a symlinked home.

## Notes for the maintainer

- Scope is deliberately minimal (8 files, +103/−8); I did **not** run `ruff format`, because the touched files were already "unformatted" by ruff's formatter on `main` and CI only runs `pytest -q` — reformatting would add unrelated noise. `ruff check` is clean on every touched file and `mypy` is clean on the changed sources.
- Optional follow-up (not included, your call): a `AGENT_REACH_CONFIG_DIR` env override for users who want the config elsewhere. `user_home()` alone fixes the reported breakage without it.

### DIFF
--- a/agent_reach/channels/github.py
+++ b/agent_reach/channels/github.py
@@ -12,6 +12,7 @@ from agent_reach.probe import probe_command
 from agent_reach.utils.paths import (
     PrivatePathError,
     read_small_text_no_follow,
+    user_home,
 )
 
 from .base import Channel
@@ -45,7 +46,7 @@ def _gh_hosts_path() -> Path:
         if app_data:
             return Path(app_data) / "GitHub CLI" / "hosts.yml"
 
-    return Path.home() / ".config" / "gh" / "hosts.yml"
+    return user_home() / ".config" / "gh" / "hosts.yml"
 
 
 def _saved_github_host_configured() -> bool:
--- a/agent_reach/channels/mcporter.py
+++ b/agent_reach/channels/mcporter.py
@@ -10,6 +10,7 @@ from pathlib import Path
 from agent_reach.utils.paths import (
     PrivatePathError,
     read_small_text_no_follow,
+    user_home,
 )
 
 _MAX_CONFIG_BYTES = 1024 * 1024
@@ -94,7 +95,7 @@ def _select_config_layers(
         return [(Path(os.path.abspath(os.fspath(expanded))), "explicit")]
 
     layers = []
-    home_base = Path.home() / ".mcporter"
+    home_base = user_home() / ".mcporter"
     for name in ("mcporter.json", "mcporter.jsonc"):
         candidate = home_base / name
         if os.path.lexists(candidate):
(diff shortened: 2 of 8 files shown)


Panniantong:
感谢反馈 symlinked HOME 场景。#575 已采用 HOME-first resolver 并保留 symlink refusal；当前方案会弱化安全门禁，因此关闭。

PR #563, opened 2026-08-03
docs(security): pin Agent Reach install source

### LINKED ISSUE #547: security(supply-chain): PyPI name `agent-reach` is unregistered while docs point users to `pip install agent-reach`
## Summary

The distribution name `agent-reach` is **not registered on PyPI**, but the README and a lot of third-party write-ups tell users to `pip install agent-reach`. Since `pyproject.toml` declares `name = "agent-reach"`, the name is currently unclaimed and available for anyone else to register. Given the project's reach, that's an open supply-chain window worth closing — either by publishing the package or by removing the PyPI-shaped instructions.

I'm not reporting a vulnerability in your code; this is about an unowned namespace that your own docs point users toward.

## Evidence

```console
$ curl -s -o /dev/null -w "%{http_code}\n" https://pypi.org/pypi/agent-reach/json
404

$ curl -s -o /dev/null -w "%{http_code}\n" https://pypi.org/simple/agent-reach/
404

$ python3 -m pip index versions agent-reach
ERROR: No matching distribution found for agent-reach

# positive control — a dependency that *is* published resolves fine
$ curl -s -o /dev/null -w "%{http_code}\n" https://pypi.org/simple/rdt-cli/
200
```

Variants `agentreach`, `agent_reach` and `xreach` are also unregistered (404).

## Why this matters

1. **`pyproject.toml` declares `name = "agent-reach"` (v1.5.0)**, so this is the canonical distribution name for the project — it just isn't claimed.
2. **The README points at it.** The install walkthrough says the CLI is installed via `pip install`, and readers reasonably complete that as `pip install agent-reach`.
3. **Contributors already assume it's published.** #375 describes "every user-facing install path (pip/pipx from main.zip, PyPI)" — treating PyPI as an existing path.
4. **Downstream coverage repeats it.** Several third-party articles and social posts about the project state `pip install agent-reach` as the install command.

If someone registers `agent-reach` on PyPI, every user who follows those instructions instead of `docs/install.md` installs their package with whatever code it contains — and it will look like the official one.

## Suggested remediation (either is fine)

**Option A — claim the name.** Register `agent-reach` on PyPI, even as a placeholder release that only points to the real install path. This closes the window permanently and is the smaller change. A [PyPI Trusted Publisher](https://docs.pypi.org/trusted-publishers/) workflow would also let releases be built from CI without long-lived tokens.

**Option B — remove the ambiguity.** If publishing to PyPI isn't planned, make the docs unambiguous that installation is only from the GitHub archive, and drop phrasings that read as a PyPI install. Worth doing in `README.md`, `docs/README_en.md` and `docs/install.md`.

Option A is probably worth doing regardless of Option B, since third-party write-ups are already out there and you can't edit those.

## Related, smaller note

`docs/install.md` installs the project itself from a mutable ref:

```bash
pipx install https://github.com/Panniantong/agent-reach/archive/main.zip
```

while the third-party dependency right below it is pinned to an exact commit:

```bash
pipx install 'git+https://github.com/public-clis/rdt-cli.git@5e4fb3720d5c174e976cd425ccc3b879d52cac66'
```

Pinning your own installs to a tag or commit would give users the same guarantee you already give them for `rdt-cli`, and would pair well with the hash-pinned lock proposed in #375.

---

Found while auditing the project before adopting it. Happy to open a PR for the docs side of Option B if that's useful.

maxitoon:
## Summary

- Fixes #547.
- Fix ambiguous install guidance for the `agent-reach` PyPI namespace reported in #547.
- Pin project installation and update examples to the `v1.5.0` GitHub archive instead of mutable `main.zip`.
- Add explicit warnings that the same-name PyPI package is unrelated to Agent Reach.
- Keep the warning consistent across the Chinese, English, Korean, and Japanese READMEs, install/update guides, `llms.txt`, generated skills, the CLI update message, the MCP diagnostic, and the integration test helper.

## Why

The repository's current package version is `1.5.0`, and the `v1.5.0` tag is available. Live PyPI metadata observed during the audit returned an unrelated `agent-reach` 0.1.0 project (`github.com/jgalea/agent-reach`). This patch does not claim maliciousness and does not attempt namespace transfer, publishing, or release configuration.

## Verification

- `git diff --check` — passed.
- `python3 -m compileall -q agent_reach` — passed.
- `bash -n test.sh` — passed.
- Dependency-free updater-message assertion — passed.
- Dependency-free MCP diagnostic assertion — passed; expected exit 1 and the actual `mcp[cli]>=1.0` dependency guidance was observed.
- Repository-wide stale-reference assertion — passed: no `archive/main.zip`; remaining bare `pip install agent-reach` text is prohibition/warning text only.
- `python3 -m pytest tests/test_cli.py -q` — 18 passed, 5 blocked before assertions by missing optional environment dependency `loguru`; no packages were installed in this read-only audit.
- `ruff` was not available in the environment.

## Scope and safety

- The contribution is limited to this single branch commit; no release, PyPI, or unrelated upstream changes were performed.
- No runtime routing behavior was changed; the Python changes update user-facing installation/error messages only.
- DCO sign-off is included.

## Review receipt

- Base: `b4d52c46c9113cb0f653d6df4cf71ebadf4930ac`
- Branch: `agent-reach-547-docs-pin`
- Commit: `188d9404ee0b8b928762997fd01ea8ad43cabf74`
- Patch SHA-256: `5f1c6dd3de487fcf65cc3210beab76d762532d8e52d6d774e7c12a254c17f8a5`

## Independent review gate

- **Status:** `PASS`
- Independent review: **PASS** on the current commit; no issues found.
- Provider smoke verification: **PASS**; the inherited-provider child started successfully.
- This PR contains no release, publishing, or PyPI configuration changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,8 @@ AI Agent 已经能帮你写代码、改文档、管项目——但你让它去�
 
 **Agent Reach 把这件事变成一句话：**
 
+> ⚠️ **安装来源提醒：** Agent Reach 不发布到 PyPI。请从下方固定版本的 GitHub 压缩包安装；不要运行 `pip install agent-reach`，PyPI 上的同名包不是本项目。
+
 ```
 帮我安装 Agent Reach：https://raw.githubusercontent.com/Panniantong/agent-reach/main/docs/install.md
 ```
@@ -135,7 +137,7 @@ AI Agent 已经能帮你写代码、改文档、管项目——但你让它去�
 <details>
 <summary>它会做什么？（点击展开）</summary>
 
-1. **安装 CLI 工具** — `pip install` 装好 `agent-reach` 命令行（自带 yt-dlp、feedparser）
+1. **安装 CLI 工具** — 从固定版本的 GitHub 压缩包安装 `agent-reach`（不要运行 `pip install agent-reach`；PyPI 同名包不是本项目）
 2. **安装系统基建** — 自动检测并安装 Node.js、gh CLI、mcporter
 3. **配置搜索引擎** — 通过 MCP 接入 Exa（免费，无需 API Key）
 4. **检测环境** — 判断是本地电脑还是服务器，给出对应的配置建议
--- a/agent_reach/cli.py
+++ b/agent_reach/cli.py
@@ -1961,7 +1961,7 @@ _UPDATE_INSTRUCTIONS = (
     "更新方式（推荐，复制这句话给你的 AI Agent，会完整更新本体+上游工具+skill）：\n"
     "  帮我更新 Agent Reach：https://raw.githubusercontent.com/Panniantong/agent-reach/main/docs/update.md\n"
     "仅更新本体（不含上游工具和 skill）：\n"
-    "  pip install --upgrade https://github.com/Panniantong/agent-reach/archive/main.zip"
+    "  pip install --upgrade https://github.com/Panniantong/agent-reach/archive/v1.5.0.zip"
 )
 
 
--- a/agent_reach/integrations/mcp_server.py
+++ b/agent_reach/integrations/mcp_server.py
@@ -28,7 +28,11 @@ except ImportError:
 
 def create_server():
     if not HAS_MCP:
-        print("MCP not installed. Install: pip install agent-reach[mcp]", file=sys.stderr)
+        print(
+            "MCP not installed. Install its dependency: "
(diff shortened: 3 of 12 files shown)


Panniantong:
感谢指出同名 PyPI 包风险。merged #575 已统一改正安装指引并增加回归门禁，因此关闭此已覆盖 PR。

PR #456, opened 2026-07-01
security: fix SSRF on WebChannel, broken API key, add prompt-injection guardrails

chrismaddern:
### 1. SSRF protection on `WebChannel.read()` (MEDIUM)

**File:** `agent_reach/channels/web.py`

`WebChannel.read()` had **zero SSRF validation** — any URL was passed directly to Jina Reader (`https://r.jina.ai/{url}`), which fetches the target URL. An agent instructed to `read http://169.254.169.254/latest/meta-data/` would fetch cloud metadata via Jina Reader, potentially leaking instance credentials.

**Fix:** Import and call `_assert_safe_public_url` from `transcribe.py` before constructing the Jina Reader URL. This reuses the existing SSRF guard that rejects private IPs, localhost, and blocked hostnames.

### 2. Fix broken API key in `transcribe_xiaoyuzhou.sh` (HIGH)

**File:** `agent_reach/scripts/transcribe_xiaoyuzhou.sh:116,140`

The curl invocations to Groq's Whisper API used `-H "Authorization: Bearer ***"` — a literal string `***` instead of `$GROQ_API_KEY`. The variable was correctly loaded (lines 31-37) and the Python polish step (line 203) used it correctly, but the two curl calls that perform the actual transcription hardcoded `***`. This made the entire xiaoyuzhou transcription feature non-functional — every transcription attempt failed with HTTP 401.

**Fix:** Replace `Bearer ***` with `Bearer "$GROQ_API_KEY"`.

### 3. Prompt-injection guardrails in SKILL.md (MEDIUM)

**File:** `agent_reach/skill/SKILL.md`

The skill file instructs the agent to fetch content from 15 platforms (tweets, Reddit posts, YouTube transcripts, web pages via Jina Reader) and feed it into the agent's context. There were **zero guardrails** instructing the agent to treat fetched content as untrusted data. A malicious tweet, Reddit post, or web page could contain text like "Ignore all previous instructions" that the agent might follow.

**Fix:** Added a prominent `⚠️ 安全规则` section to SKILL.md warning the agent that all fetched content is UNTRUSTED DATA and must never be treated as instructions.

## Test Results

All 196 existing tests pass:
```
============================= 196 passed in 8.02s ==============================
```

## Security Audit Context

[...]

Issue #444, opened 2026-06-28
SSRF vulnerability in transcribe() — CWE-918 (Server-Side Request Forgery)


nobugpal:
## Summary

**CWE-918: Server-Side Request Forgery**

A Server-Side Request Forgery vulnerability exists in `agent_reach/transcribe.py:77-98` (`download_audio()`). The function passes user-supplied URLs directly to `yt-dlp` with zero validation, allowing an attacker to make HTTP requests to arbitrary internal hosts, including private network services and cloud metadata endpoints.

## Disclosure Note

The repository security policy recommends using GitHub Private Vulnerability Reporting (PVR). However, the PVR link returns a 404 error — the feature is not enabled on this repository. Per the policy's scope, SSRF is listed as "in scope". Since the private channel is unavailable, this public issue is the only way to report the vulnerability to the maintainer.

---

## Vulnerability Details

**Root cause**: `agent_reach/transcribe.py:77-98` — `download_audio()` passes the `url` parameter directly to `yt-dlp` without any validation:
```python
def download_audio(url: str, out_dir: Path) -> Path:
    _run(["yt-dlp", "-x", "--audio-format", "m4a", url])  # url = raw attacker input
```

The URL is accepted through three paths:
1. CLI: `agent-reach transcribe <url>`
2. Python API: `agent_reach.transcribe.transcribe(url)`
3. Agent tool: `YouTubeChannel.transcribe(url)` — exploitable via prompt injection

**No validation exists**: no scheme allowlist, no private-IP blocklist, no SSRF protection of any kind.

## Impact

Confirmed via Docker-isolated retest (full attacker chain, L4 evidence):

| Scenario | Impact |
|---|---|
| Internal service discovery | Attacker probes live hosts/ports behind firewall |
| Data exfiltration | yt-dlp downloaded internal WAV file (1644 bytes) to attacker disk |
| Cloud metadata credential theft | AWS/GCP/Azure IAM credentials accessible via 169.254.169.254 |
| Internal DB discovery | Unauthenticated database admin interfaces leak connection strings |

**CVSS v3.1**: 6.8 (Medium-High) — `CVSS:3.1/AV:N/AC:L/PR:N/UI:R/S:C/C:H/I:N/A:N`

## Fix Available

A fix PR is submitted alongside this issue: **PR #443**

[...]

PR #443, opened 2026-06-28
fix: add SSRF validation to download_audio() before passing URL to yt-dlp

nobugpal:
## Summary

Adds SSRF (Server-Side Request Forgery, CWE-918) protection to the `transcribe()` function. The `download_audio()` function previously passed user-supplied URLs directly to `yt-dlp` with **zero validation**, allowing an attacker to instrument HTTP requests to arbitrary internal/private hosts.

## Vulnerability

**CWE-918: Server-Side Request Forgery**

Affected function: `agent_reach/transcribe.py:77-98` — `download_audio()`

An attacker with access to the CLI or API could run:
```
agent-reach transcribe http://169.254.169.254/latest/meta-data/
agent-reach transcribe http://192.168.1.1/admin
```
And yt-dlp would make requests to those internal hosts.

## Fix

Added `_check_ssrf(url)` that validates URLs before passing them to yt-dlp:
- Blocks localhost (127.0.0.1, ::1, localhost, 0.0.0.0)
- Blocks private IP ranges (10.x.x.x, 172.16-31.x.x, 192.168.x.x)
- Blocks cloud metadata endpoints (169.254.169.254, etc.)
- Blocks non-HTTP/S URL schemes
- Resolves DNS and checks resolved IPs against private ranges
- DNS failures are tolerated (yt-dlp handles them)

## Verification

- 17/17 SSRF test cases pass (4 positive, 13 negative)
- All 157 pre-existing tests continue to pass (5 pre-existing failures are unrelated loguru import errors)
- Docker retest confirmed: SSRF blocked after fix, functionality preserved for public URLs

## Related

This PR addresses a privately reported security vulnerability. See the corresponding security advisory for full details.

PR #564, opened 2026-08-04
fix(transcribe): block shorthand IPv4 spellings of internal hosts

SEPURI-SAI-KRISHNA:
## Problem

`_assert_safe_public_url()` in `agent_reach/transcribe.py` is the SSRF guard added by #443 (closing #444). It rejects internal targets by parsing the host with `ipaddress.ip_address()` — which only accepts the **canonical dotted-quad** form.

The C resolver behind `yt-dlp` accepts the whole `inet_aton` grammar. So every non-canonical spelling of an internal address walks straight past the check:

| URL passed to `transcribe` | Guard says | Actually reaches |
|---|---|---|
| `http://127.1/` | allowed | `127.0.0.1` |
| `http://127.0.1/` | allowed | `127.0.0.1` |
| `http://2130706433/` | allowed | `127.0.0.1` |
| `http://0x7f000001/` | allowed | `127.0.0.1` |
| `http://0177.0.0.1/` | allowed | `127.0.0.1` |
| `http://0/` | allowed | `0.0.0.0` |
| `http://192.168.1/` | allowed | `192.168.0.1` |
| `http://2852039166/` | allowed | **`169.254.169.254`** |
| `http://0xA9FEA9FE/` | allowed | **`169.254.169.254`** |

The last two reach the cloud instance-metadata endpoint — the exact target the guard names in its own `_BLOCKED_HOSTS` neighbourhood (`metadata.google.internal`). `http://169.254.169.254/` is blocked; `http://0xA9FEA9FE/` is not.

Reproduce on `main`:

```python
from agent_reach.transcribe import _assert_safe_public_url
_assert_safe_public_url("http://0xA9FEA9FE/")   # returns — no exception
```

```python
import socket
socket.getaddrinfo("0xA9FEA9FE", 80)[0][4]      # ('169.254.169.254', 80)
```

This matters because `transcribe(source)` takes its URL from `agent-reach transcribe <source>` and from `YouTubeChannel.transcribe()`, i.e. from an agent or from scraped content — precisely the untrusted-input path #444 was filed about.

## Fix

Parse the host with the same grammar the resolver uses, then run the existing private/loopback/link-local/reserved/multicast/unspecified check against the canonicalised address.

`socket.inet_aton` is the same libc parser `getaddrinfo` uses for IPv4 literals, and it rejects real hostnames (`example.com`, `localhost`, `1.2.3.4.5` → `OSError`), so it is a pure literal parse.

**No DNS lookup is added.** The existing `test_does_not_dns_resolve_public_hostnames` contract is preserved, and a new test asserts the shorthand path stays DNS-free by making `socket.getaddrinfo` raise.

Public literal addresses are unaffected — including `http://010.010.010.010/`, which is octal for the public `8.8.8.8` and stays allowed.

## Tests

`tests/test_transcribe.py`, in the existing `TestDownloadAudioSafety` class:

- `test_rejects_shorthand_ipv4_spellings_of_internal_hosts` — 10 parametrised spellings, each asserting `yt-dlp` is never invoked
- `test_shorthand_ipv4_check_stays_dns_free` — fails if the new parsing resolves a name
- `test_allows_public_literal_addresses` — regression guard for `1.1.1.1`, `8.8.8.8`, `010.010.010.010`

`pytest -q`: 442 passed (was 428). `ruff check` clean on both touched files.

## Scope

Deliberately narrow — one function, no behaviour change for any valid input. It does not touch the separate `WebChannel` SSRF discussion in #456/#370, and does not add redirect-following or DNS-rebinding protection; those remain out of scope for this guard.

### DIFF
--- a/agent_reach/transcribe.py
+++ b/agent_reach/transcribe.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 import ipaddress
 import math
 import shutil
+import socket
 import subprocess
 import tempfile
 from pathlib import Path
@@ -159,10 +160,30 @@ def _run(cmd: List[str], timeout: int = 600) -> None:
         )
 
 
-def _is_private_ip(value: str) -> bool:
+def _literal_ip(host: str):
+    """Return the address a literal host denotes, or None for a real hostname.
+
+    ``ipaddress`` only accepts the canonical dotted-quad form, but the C
+    resolver behind yt-dlp accepts the whole ``inet_aton`` grammar: ``127.1``,
+    ``2130706433``, ``0x7f000001`` and ``0177.0.0.1`` all reach 127.0.0.1, and
+    ``0xA9FEA9FE`` reaches the cloud metadata endpoint. Parsing with the same
+    grammar keeps those shorthands from slipping past the private-address
+    check. This is literal parsing only — no name is resolved here.
+    """
     try:
-        ip = ipaddress.ip_address(value)
+        return ipaddress.ip_address(host)
     except ValueError:
+        pass
+    try:
+        packed = socket.inet_aton(host)
+    except OSError:
+        return None
+    return ipaddress.IPv4Address(packed)
+
+
+def _is_private_ip(value: str) -> bool:
+    ip = _literal_ip(value)
+    if ip is None:
         return False
     return any(
         (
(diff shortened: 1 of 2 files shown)


Panniantong:
感谢发现 shorthand IPv4 SSRF 绕过。你的原始作者提交已保留在 merged #577 中，并扩展覆盖 Unicode IDNA、percent-encoded host 与反斜杠 authority，完整 CI 已通过，因此关闭原 PR。

PR #571, opened 2026-08-05
test(exa_search): add unit tests for the ExaSearchChannel doctor branches

nyxst4ck:
## What

Dedicated offline tests for the `exa_search` channel — the channel had no dedicated test file. 6 tests covering:

- `can_handle`: always `False` (search-only channel), pinned so a future URL route can't slip in silently
- `check()` all five branches:
  - mcporter missing → `off` with both install prescriptions (`npm install -g mcporter` + `mcporter config add exa`)
  - unreadable mcporter config → `error` (and stale `active_backend` cleared)
  - `exa` present in config → `warn`, never `ok` — Doctor never probes the remote MCP, so config presence alone must not be sold as usable
  - editor imports left unexpanded → `warn` (credential-read boundary)
  - mcporter installed but Exa unconfigured → `off` with the exact add command

## How

`inspect_mcporter_config` and `shutil.which` are stubbed at the channel module, so every branch runs offline and deterministic — same approach as the merged channel-test series.

## Test plan

- `pytest tests/test_exa_search_channel.py -v` → 6 passed
- Full suite: 416 passed (the 10 failures in `test_private_file_writes.py` are pre-existing Windows symlink-permission failures on my machine, untouched by this PR)

Follow-up to #331 — continues the dedicated channel coverage series after rss (#360), github (#361), web (#363), reddit (#364), xueqiu (#365), v2ex (#366) and youtube (#367).

### DIFF
--- a/tests/test_exa_search_channel.py
+++ b/tests/test_exa_search_channel.py
@@ -1,0 +1,109 @@
+# -*- coding: utf-8 -*-
+"""Dedicated tests for the ``exa_search`` channel.
+
+Exa is a search-only channel (``can_handle`` is always False) that rides
+mcporter + the remote Exa MCP. Its ``check()`` has five outcomes: mcporter
+missing (off with install prescription), unreadable mcporter config
+(error), ``exa`` present in config (warn — remote connectivity is never
+probed by Doctor), editor imports left unexpanded (warn — credential-read
+boundary), and mcporter installed but Exa unconfigured (off with the exact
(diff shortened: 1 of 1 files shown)


Panniantong:
感谢补充 ExaSearchChannel Doctor 分支测试。merged #577 已在 tests/test_channels.py 覆盖 mcporter 缺失、配置错误、精确 server name、editor imports 与未配置路径，并同步了当前 Exa 工具契约，因此关闭此重复测试 PR。

PR #502, opened 2026-07-13
feat: add FirecrawlChannel for deep web scraping and search

nankingjing:
Add a new Firecrawl-based channel for Agent-Reach that provides:

- **read()**: Scrape any URL to clean markdown via Firecrawl API
- **crawl()**: Deep crawl websites with configurable depth/limits
- **search()**: Web search with structured results
- **extract()**: AI-powered structured data extraction

The channel is **tier 1** (needs free API key from firecrawl.dev).
Gracefully reports missing dependencies and API key status in check().

### Changed files:
- agent_reach/channels/firecrawl.py - New FirecrawlChannel
- agent_reach/channels/__init__.py - Register FirecrawlChannel
- agent_reach/config.py - Add firecrawl to FEATURE_REQUIREMENTS
- tests/test_firecrawl_channel.py - 25 unit tests

Tests: **25/25 passing** + all contract tests pass.

PR #596, opened 2026-08-09
feat(web): configurable Firecrawl backend with Jina fallback

LAP87:
## Summary

Makes the web channel's Firecrawl backend **configurable and optional** — the self-hosted Firecrawl instance is no longer hardcoded. Firecrawl remains the preferred backend, with Jina Reader as the always-available fallback.

## Changes

**`agent_reach/channels/web.py`**
- `backends = ["Firecrawl", "Jina Reader"]` — Firecrawl preferred, Jina fallback, following the `base.py` backend-routing pattern
- Base URL configurable via `FIRECRAWL_URL` env or `firecrawl_url` config key (default `http://localhost:13002`)
- Optional API key via `FIRECRAWL_API_KEY` / `firecrawl_api_key` — sent as Bearer token on the wire, never logged
- `web_backend` / `WEB_BACKEND` override reorders candidates via `ordered_backends(config)`
- `check()` probes Firecrawl → `ok`/active=Firecrawl when reachable; `warn`/active=Jina Reader when down; override to Jina skips the probe (zero overhead)
- `read()` iterates `ordered_backends`: **transport errors** (`OSError` family) fall back to the next backend; **content errors** (oversize, antibot, Firecrawl `success:false`/bad JSON) propagate immediately
- Preserved `normalize_public_http_url`, 5 MiB limit, and Jina antibot detection

**`tests/test_web_channel.py`** — 56 tests covering both backends, config/env resolution, override, fallback semantics, and error propagation. All mocked, no network.

## Testing

- `pytest tests/ -q` → **601 passed, 16 subtests passed**
- Live smoke: `check()` → `ok` with `active_backend='Firecrawl'` against a running instance; `read()` returns markdown via Firecrawl; `web_backend='Jina Reader'` override routes via Jina

## Notes

- Uses stdlib `urllib` only — no new dependencies (glue-layer rule)
- API key is never sent unless configured; default URL points at localhost so nothing leaks by default

### DIFF
--- a/agent_reach/channels/web.py
+++ b/agent_reach/channels/web.py
@@ -1,7 +1,26 @@
 # -*- coding: utf-8 -*-
-"""Web — any URL via Jina Reader. Always available."""
+"""Web — any URL via a configurable backend (Firecrawl preferred, Jina fallback).
 
+Firecrawl is a self-hostable reader. Its base URL comes from the env var
+``FIRECRAWL_URL`` or the config key ``firecrawl_url`` (default
+``http://localhost:13002``) so a user's own install is honored. Jina Reader
+remains the always-available public fallback.
+
+Backend routing follows ``base.py``: ``backends`` is an ordered candidate
+list (``backends[0]`` preferred); ``ordered_backends(config)`` honors a user
+override via ``web_backend`` / ``WEB_BACKEND``.
+
+Fallback semantics: a *transport* failure (network error, timeout, HTTP
+error status) on Firecrawl falls back to Jina. A *content* failure (oversize
+response, antibot page, Firecrawl scrape error) propagates as an error and
+does NOT trigger the fallback — the caller sees the real failure.
+"""
+
+import json
+import os
 import urllib.request
+
+from loguru import logger
 
 from agent_reach.utils.url import normalize_public_http_url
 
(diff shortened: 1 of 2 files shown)


yoouth:
Hi @LAP87, @Panniantong — I've been working on Firecrawl support in a fork and want to align direction before proposing anything, given #502's closure.

My approach differs from this PR's: a standalone check-only `FirecrawlChannel` (mirroring `exa_search.py`) that inspects mcporter config and never wraps the SDK — agents call the Firecrawl MCP tools directly via mcporter. To address the "no real API verification" concern from #502, I'm adding an opt-in `doctor --probe` that makes one minimal authenticated call and reports ok/error, keeping default `doctor` runs free of remote calls.

Two questions: (1) Is the channel-vs-web.py-backend question settled — should Firecrawl live inside the `web` channel as this PR does, or is a standalone channel acceptable for the search/crawl surface that doesn't fit `web.read()`? (2) Does an opt-in probe satisfy the verification positioning, or should `check()` probe by default as this PR's does?

Happy to converge with this PR rather than compete — the mcporter-based routing and the self-hosted `FIRECRAWL_URL` backend here look complementary, not conflicting.

PR #644, opened 2026-08-21
fix(config): accept truthy-but-valid values in is_configured()

abhijeet117:
## Summary
Config.is_configured() rejected valid config values like 0, '', and [] because it used Python truthiness (all(...)) instead of checking is not None. Combined with Config.get() also using if env_val: (line 166), empty-string env vars like EXA_API_KEY="" were silently ignored.

## What changed
agent_reach/config.py — Changed all(self.get(k) for k in required) to all(self.get(k) is not None for k in required) in is_configured().
tests/test_config.py — Added 4 test cases covering empty string, zero, empty list, and None.

## Testing
- test_is_configured_accepts_empty_string — passes
- test_is_configured_accepts_zero — passes
- test_is_configured_accepts_empty_list — passes
- test_is_configured (existing) — still returns False when key is absent

### DIFF
--- a/agent_reach/config.py
+++ b/agent_reach/config.py
@@ -199,7 +199,7 @@ class Config:
     def is_configured(self, feature: str) -> bool:
         """Check if a feature has all required config."""
         required = self.FEATURE_REQUIREMENTS.get(feature, [])
-        return all(self.get(k) for k in required)
+        return all(self.get(k) is not None for k in required)
 
     def get_configured_features(self) -> dict:
         """Return status of all optional features."""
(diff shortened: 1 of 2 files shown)


abhijeet117:
Closing this one. On a second look the premise does not hold.

Every key in `FEATURE_REQUIREMENTS` is a credential (`exa_api_key`, `groq_api_key`, `twitter_auth_token`, `github_token`), and `is_configured()` gates whether a provider is reported as usable, for example in the transcription check in `channels/youtube.py`. An empty API key is genuinely not configured, so the existing truthiness check is the correct behavior. Treating `""` as configured would make `doctor` report a provider as available when any call with that key would fail.

The `0` and `[]` cases cannot occur either, since none of the required keys are numeric or list valued. Sorry for the noise.

PR #687, opened 2026-09-11
fix(paths): tolerate chmod denial in restricted environments

Oscar-Williams:
## Summary

Fixes the failure mode reported in #684. `make_private_dir()` no longer calls `fchmod()` when the directory is already `0700`, avoiding the denied no-op syscall seen in restricted sandboxes. When tightening is needed but the sandbox denies `fchmod`, the helper emits a `RuntimeWarning` and continues; symlink checks, fd cleanup, Windows behavior, and private-file handling are unchanged.

Added regression tests for the skip, successful hardening, and `PermissionError` paths.

## Test Plan

- `ruff check agent_reach tests` — passed.
- `mypy agent_reach` — passed.
- Focused path/config/private-file tests — 45 passed, 14 skipped, 1 deselected.
- Full Windows suite — 551 passed, 28 skipped, 10 platform failures; the same 10 failures reproduce on the unmodified base.
- Linux `make_private_dir` smoke test — passed.


Oscar-Williams:
Friendly ping — this fixes `make_private_dir` crashing `doctor` in restricted environments (the unconditional `chmod`). It's been open a while with no feedback; let me know if anything needs changing, or if it's out of scope I'm happy to close it.

